PR #124, opened 2014-12-15
[fixes #83982776] Improve reporting output for dependencies that have no...

mhess:
...t been installed.  Rename 'other' to 'unknown' for packages for which we cannot determine the license.


mainej:
Switched name of 'other' to 'unknown' in 044a89755fc.  @mhess I remembered to set you as the author on the 'missing' change, but not on the 'other'->'unknown' change. Sorry about that.  Calling this done.

PR #132, opened 2015-01-11
Compost cucumber

mainej:
This commit replaces the cucumber tests with integration-style tests written in rspec.  

Before, in order to write integration tests, a developer had to understand both the ruby-based testing DSL and how to wire that into Cucumber.  The testing DSL was already reasonably "human readable".  So, Cucumber was an additional layer of indirection without any tangible benefit.  Now, without Cucumber, the tests provide much better examples of how to actually use `license_finder`.

This patch also splits up the testing DSL into independent components, and tries to make the test code even more readable.  In the process, I clarified and focused the integration tests.  Together, these changes shave about 40% (1.5 minutes) off the suite locally and about 30% (16 minutes) off the Travis build.


flavorjones:
Bonus points for branch naming.

PR #133, opened 2015-01-16
Forbid approval of dependencies with exclusively blacklisted licenses

mainej:
This doesn't do anything to highlight blacklisted packages.  Instead it just refuses to show them as approved, regardless of the whitelist or manual approvals.  Easy enough to change if that's important.

A lot of this code is an almost exact copy of the whitelist code.  Would love feedback about how to improve that.

[#84478498]


erd:
Rebased and committed in f986a35b2f92dc231990850e785f3297cbd569c0

PR #136, opened 2015-02-11
Wired rebar (erlang package manager) to support discovery of licenses in...

gregdurham:
... erlang projects

Testing for core rebar functionality

Updated the README to include rebar.


gregdurham:
It appears there is an issue with the python test, I am not sure the correct path to install erlang/rebar into the vm for testing. If anyone has any ideas I would love to hear them! Thanks for the awesome tool!

mainej:
Thanks for putting this together @gregdurham. To get the tests running, I think you'll have to configure .travis.yml much like we do for gradle.  That is, wget rebar and make sure it's somewhere on the PATH.  Our travis vm is ruby, so most other things need to be manually installed.  I don't know whether erlang itself is installed on the vm, so if not, there's probably more work to do.

I'm not sure what's going on with the python tests.  There's a chance NPM was down.  But we can figure that out after the erlang tests are passing.

features/features/package_managers/rebar_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'feature_helper'
+
+describe "Rebar Dependencies" do
+  # As a Node developer
+  # I want to be able to manage NPM dependencies
```
mainej: Should be "As an Erlang developer / I want to be able to manage rebar dependencies".

features/features/package_managers/rebar_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  let(:erlang_developer) { LicenseFinder::TestingDSL::User.new }
+
+  specify "are shown in reports" do
+    LicenseFinder::TestingDSL::RebarProject.create
+    erlang_developer.run_license_finder
+    expect(erlang_developer).to be_seeing_line "sqerl, \"BRANCH: master\", \"Apache 2.0\""
```
mainej: This is the only mention of `sqerl` ... is `sqerl` a dependency of all `rebar`-managed projects?  Or is it missing from `spec/fixtures/all_pms/rebar.config`?

lib/license_finder/package_managers/rebar.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    def capture(command)
+      [`#{command}`, $?.success?]
+    end
+
+    def package_path
+      project_path.join('rebar.config')
+    end
+
+    def lockfile_path
```
mainej: Is this needed?

lib/license_finder/package_managers/rebar.rb
```diff
@@ -0,0 +1,60 @@
...
+              "name" => dependency[0],
+              "version" => "#{dependency[1]}: #{dependency[2]}",
+              "homepage" => dependency[3]
+            }
+          end
+        end
+      else
+        raise "Command #{command} failed to execute: #{output}"
+      end
+      JSON(dependencies)
```
mainej: `dependencies` doesn't need to be serialized into JSON if it will be immediately deseralized in `current_packages`.

lib/license_finder/package_managers/rebar.rb
```diff
@@ -0,0 +1,60 @@
+require 'json'
+
+module LicenseFinder
+  class Rebar < PackageManager
+    def initialize(options={})
+      super
+      @command = options[:rebar_command] || "rebar"
+      @deps_path = options[:rebar_deps_dir] || "deps"
```
mainej: If you change this to 
```ruby
@deps_path = Pathname(options[:rebar_deps_dir] || "deps")
```
then below you can write
```ruby
@deps_path.join(dep["name"])
```
which is slightly shorter... we try to use Pathname if we can.

lib/license_finder/package_managers/rebar_package.rb
```diff
@@ -0,0 +1,15 @@
+module LicenseFinder
+  class RebarPackage < Package
+    def initialize(name, version, install_path, dep, options={})
+      super(
+        name,
+        version,
+        options.merge(
+          description: name,
```
mainej: It's OK to leave description blank if you can't fetch any more descriptive text from rebar.

features/features/package_managers/rebar_spec.rb
gregdurham: Thank you for catching that!

features/features/package_managers/rebar_spec.rb
gregdurham: I based this off of how other package_managers were implemented, it was just a sample package to ensure that rebar support was working in the integration tests.

lib/license_finder/package_managers/rebar.rb
gregdurham: Nope, can be removed.

gregdurham:
Thank you so much for all the tips, Ill get this updated!

features/features/package_managers/rebar_spec.rb
gregdurham: There is a missing file 'features/fixtures/rebar.config' which is a sample rebar config file.

gregdurham:
Thanks @mainej! I am going to try and put something together for the travis piece, the biggest issue is that build times will drastically increase due rebar requiring erlang, which will most likely need to be built from source on these boxes. The only other solution is to switch them to erlang machines (do they include Ruby?) the issue there is that the script param expects a make command, which is not needed for this project.

gregdurham:
@mainej, rebar is now part of the build, was significantly easier than I expected to add it. All of the Erlang tests are now passing. The only test failing now is the python test from earlier. Thanks again!

mainej:
@gregdurham thanks for making those tweaks and getting the rebar integration test running.  I definitely want to get this merged, but I'm on vacation so it will have to wait until early March, unless @flavorjones can get to it.  The pip integration test is still broken, and perhaps the gradle test under JRuby, but I think those are unrelated to this PR.

If you have time and interest the `rebar_command` and `rebar_deps_dir` config options need to be wired up before they will work:
* [ ] They should be settable as CLI options, much like [`gradle_command`](https://github.com/pivotal/LicenseFinder/blob/master/lib/license_finder/cli/main.rb#L18).
* [ ] They should be [passed into](https://github.com/pivotal/LicenseFinder/blob/master/lib/license_finder/cli/base.rb#L16-L25) LicenseFinder::Core, 
* [ ] [handled](https://github.com/pivotal/LicenseFinder/blob/master/lib/license_finder/configuration.rb#L14-L16) (and maybe, defaulted) in LicenseFinder::Configuration, 
* [ ] and [passed through](https://github.com/pivotal/LicenseFinder/blob/master/lib/license_finder/core.rb#L57-L65) to LicenseFinder::PackageManager.
* [ ] They should be documented in the [configuration](https://github.com/pivotal/LicenseFinder#configuration) section of the README.

If you don't have time to get to this, I'll do it soon after merging.

gregdurham:
Thank you for the very detailed instructions, the latest should include the changes to make it wired into the configuration. Also, I think for future use this could be documented on the wiki, as it was very helpful.

PR #138, opened 2015-03-02
Pip tests are broken on Travis

mainej:
Travis upgraded pip (6.0.7?), and now the integration tests fail.  This branch is a series of experiments to try to get the tests passing.

The core problem is that we no longer have permissions for the system python/pip install directory.  I think if we could get everything to run under virtualenv, that would fix it.  

If that fails, we could cheat.  Instead of testing that license_finder reports a custom package, we could test that it reports one of the packages installed on every system (like rdoc is installed for all rubies).  After all, our job is not to install packages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,3 +26,7 @@ before_install:
   - .travis/install_gradle.sh
   - .travis/install_rebar.sh
   - .travis/install_bower.sh
+  - .travis/install_virtualenv.sh
+
+before_script:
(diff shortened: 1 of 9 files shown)


mainej:
This first commit is the cheat mentioned in the PR description... just seeing if it works.

mainej:
Finally got the tests to run in virtualenv on Travis.  Gonna go with this approach, instead of cheating.

PR #142, opened 2015-03-10
Support for GPLv3 in license_finder

gregdurham:


mainej:
Merged. Will be released as part of version 2.0.2.  Thanks!

PR #144, opened 2015-03-26
Identify SimplifiedBSD in rubyzip: BSD 2-Clause

eikes:
Rubyzip (and possibly others) use "BSD 2-Clause" as the license name:

https://github.com/rubyzip/rubyzip/blob/master/rubyzip.gemspec


mainej:
Thanks!

Issue #113, opened 2014-09-22
Support node dependency groups


taavo:
Currently LicenseFinder is [punting](https://github.com/pivotal/LicenseFinder/blob/ad14c838e0d34c3941c3e973820b977206464b10/lib/license_finder/package_managers/npm_package.rb#L32) on the existence of npm dependency groups (e.g. "dependencies" vs "devDependencies"). It would be great to be able to exclude npm devDependencies the same way you can exclude bundler groups.

cc @nertzy

mainej:
Fixed by #152.

nkoterba:
So perhaps I'm missing something but I don't see how #152 addresses this question.

I'd like to EXCLUDE `devDependencies` from LicenseFinder when crawling my npm-based directories.

However, the "fix" in #152 appear to always include both `dependencies` and `devDependencies`.

Am I missing some configuration item or command line argument?

mainej:
@nkoterba Have you run `license_finder ignored_groups add devDependencies` followed by `license_finder report`? If that's not working, you've found a bug.  A minimal package.json which reproduces the problem would help.

nkoterba:
@mainej Just verified that adding `devDependencies` as an ignored group does **not** filter out dev packages.

Keep in mind this is with npm/node.  According to the README.md, `ignored_groups` is only supported with Bundler.

Not sure if the documentation is correct or outdated.

Here's a simple package.json file that should show the issues:
```javascript
{
  "name": "test_ignored",
  "version": "1.0.0",
  "description": "Testing Ignored Groups with License_Finder",
  "dependencies": {
    "angular": "^1.4.3"
  },
  "devDependencies": {
    "gulp-rename": "^1.2.2"
  },
  "license": "ISC"
}
```

Here's what my `dependency_decisions.yml` file looks like after I ran the command you suggested above:
```javascript
---
- - :ignore_group
  - devDependencies
  - :who: 
    :why: 
    :when: 2015-07-30 01:49:51.659344000 Z
```

And here's my output:
```javascript
nk-mbp:webclient a$ license_finder
..........Dependencies that need approval:
angular, 1.4.3, MIT
gulp-rename, 1.2.2, MIT
```

If working, I believe we should only see `angular` there.

mainej:
hmm.... @LukeWinikates any thoughts on this?

mainej:
[...]

PR #152, opened 2015-05-13
support npm dependency groups

LukeWinikates:
Responding to issue #113. I favor using NPM's own terms of `dependencies` and `devDependencies` rather than mapping these to other terms.

### DIFF
--- a/lib/license_finder/package_managers/npm.rb
+++ b/lib/license_finder/package_managers/npm.rb
@@ -5,18 +5,41 @@ module LicenseFinder
     DEPENDENCY_GROUPS = ["dependencies", "devDependencies"]
 
     def current_packages
-      json = npm_json
-      dependencies = DEPENDENCY_GROUPS
-        .map { |g| (json[g] || {}).values }
-        .flatten(1)
-        .reject{ |d| d.is_a?(String) }
-
-      pkgs = {} # name => spec
-      dependencies.each { |d| recursive_dependencies(d, pkgs) }
-      pkgs.values.map { |d| NpmPackage.new(d, logger: logger) }
+      packages = {}
+      direct_dependencies.each do |dep|
+        group_name = dep[:group]
+        walk_dependency_tree(dep[:name]) do |dependency|
+          package_id = dependency["name"]
+          packages[package_id] ||= NpmPackage.new(dependency, logger: logger)
+          packages[package_id].groups << group_name unless packages[package_id].groups.include?(group_name)
+        end
+      end
+      packages.values
     end
 
     private
+
+    def direct_dependencies
+      package_json = JSON.parse(File.read(package_path))
+      DEPENDENCY_GROUPS.map do |group|
+        package_json.fetch(group, {}).keys.map do |dependency|
+          {
+            group: group,
+            name: dependency
+          }
+        end
+      end.flatten
+    end
+
+    def walk_dependency_tree(dependency, &block)
+      @json ||= npm_json
+      deps = @json.fetch("dependencies", {}).reject { |_,d| d.is_a?(String) }
+      current_dep = deps[dependency]
+      block.call(current_dep) if current_dep
+      recursive_dependencies(current_dep) do |d|
+        block.call(d)
+      end
+    end
 
     def npm_json
       command = "npm list --json --long"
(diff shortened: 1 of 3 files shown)


mainej:
I'd like to get something like this merged in... Is there a reason you closed this PR @LukeWinikates?

LukeWinikates:
I'd like to take another look at it-- I think the pull request was
premature. Specifically, I thought the key to supporting dependency groups
was populating the `groups` property on the NpmPackage objects themselves,
but looking at the code for Bundler dependencies it looks like the idea is
that `current_packages` does the filtering and the `groups` property is
only there because we get it for free from Bundler. If that's the case, the
change in this PR is actually not going to help.

I made another go at it this past week and came up with a simpler
implementation, but hit a different snag. The `npm list --json --long`
command was listing both `dependencies` and `devDependencies` under the
`dependencies` field. My pair and I ran out of time before we could dig
deeper than that, but it seems possible that `npm list` isn't producing
output we can use.

2015-05-25 20:27 GMT-07:00 Jacob Maine <notifications@github.com>:

mainej:
I wouldn't look to the Bundler code for guidance... that's the oldest part of the codebase, and does things sort of strangely.  You're right, it filters out dependencies in the ignored groups before making `BundlerPackage`s, but I've thought about removing that.  It doesn't need that responsibility because `DecisionApplier#acknowledged` also filters ignored groups, and it's supposed to be the real authority.  As you originally suspected, the real priority is getting `NpmPackage#groups` populated. `DecisionApplier` can take it from there.

That second problem with `devDependencies` being listed under `dependencies` sounds tricker.

mainej:
Moved the responsibility for ignoring groups out of `Bundler` as of 20d98aa12f.  It should be clearer now that `PackageManager`s don't have to do anything except make sure`Package#groups` are set.

LukeWinikates:
Wow. In that case the PR is probably *close* to good enough.

The `npm ls` output seems to be good for getting the recursive
dependencies, but I'm feeling like parsing `package.json` ourselves to get
the `dependencies` and `devDependencies` is the way to go to finish this
off.


2015-05-26 22:17 GMT-07:00 Jacob Maine <notifications@github.com>:

mainej:
After a bit more thought... I need to put the group filtering back into Bundler.  Taking it out was a bigger change than I originally thought.  The behavior originally was "ignore dependencies in excluded groups *and* their dependencies (which have no group)".  Now it's only "ignore dependencies in excluded groups", meaning recursive dev dependencies will show up.  Sorry for jumping the gun on this.

LukeWinikates:
Ok, this works now. I tried it on a couple of different repos:

[typicode/json-server](http://github.com/typicode/json-server)
[Polymer/polymer](http://github.com/Polymer/polymer)
[purifycss/purifycss](http://github.com/purifycss/purifycss)

To solve the `npm list` problem, this commit points LicenseFinder first at the `package.json` to get the list of direct dependencies, then at the `npm list` output to get the full dependency graph for each dependency.

LukeWinikates:
Travis build seems to be failing on master. Can't tell if there's an issue with anything specific to this change set.

mainej:
It looks like the tests are broken because of other changes on master... I've asked that some of the other developers look into that.

LukeWinikates:
Ok, getting a better understanding of the state of ignored groups. I 

It wouldn't be a big deal to switch to `(DEPENDENCY_GROUPS - options[:ignored_groups])` and drop the code that sets the groups field on the package, though I'd want to find a place for that to live outside of `Bundler` or `NPM` since it would be duplicated between them.

#156 is a separate pull request about making ignored groups an `all?` instead of an `any?`, with the caveat of still returning `false` if the dependency isn't a member of any explicit group. I'm guessing this is how Bundler dependencies without any explicit group appear. I was assuming that this was the current behavior. I appreciate the correction.

I'll update the PR with some more variety in the test cases, per your suggestion.

mainej:
Thanks for #156... I'll get that merged when the build is fixed.  And when it's merged, there will be no need for `(DEPENDENCY_GROUPS - options[:ignored_groups])`, correct?  That is, unlike Bundler, NPM will propagate parent groups to their children.  Therefore, children without a parent in an "included" group will be filtered out by DecisionApplier, and don't need to be filtered out by the PackageManger.

Maybe we could make Bundler do the same propagation... so that recursive dependencies would have the groups of their parents, instead of an empty set of groups.  Then we could move ignored_groups out of Bundler for good.

LukeWinikates:
Yep, propagating the parent groups is the goal. There won't be any need for `(DEPENDENCY_GROUPS - options[:ignored_groups])`. Part of me likes filtering out the dependency groups up front like Bundler does, rather than building up the whole dependency list and filtering out later. But having the ignoring behavior pulled out of individual package managers and up into a higher level definitely separates concerns better and I think it's the better way to go.

I haven't taken a second look at the Bundler code. That could be a fun project.

mainej:
Merged. Thanks!

PR #156, opened 2015-06-19
make DecisionApplier#ignored? return false when some of a package's explicit groups are not ignored

LukeWinikates:
see #152


mainej:
:+1: I just checked... you're correct, recursive bundler dependencies are not part of any group.  So, good call on adding `package.groups.any?`.

I'll get this merged when the build is fixed.

mainej:
Merged, though the build is still slightly broken.

Issue #155, opened 2015-06-17
Nuget support


LukeWinikates:
Nuget is the package manager for .Net applications.

Two things about Nuget support will be interesting:
1) The license for a package is indicated by a `licenseUrl` field ([see this page](http://docs.nuget.org/create/nuspec-reference)), not a file in the repo or an obvious shorthand like "MIT", so identifying licenses may be tricky.
2) Nuget doesn't distinguish between different types of dependencies, so there's not a first class concept that maps cleanly to the idea of test or development dependencies from Bundler et. al. Instead there's a convention of breaking test code out into its own "assembly" (package) with its own dependency list, so you'll have a structure like:

```
+-- MyApp.sln
|   +-- Api/
|       +-- Api.csproj
|       +-- package.config <--- list of dependencies for the Api
|   +-- Api.Test/
|       +---Api.Test.csproj
|       +---package.config <-- list of dependencies for the Api Tests
```

To support ignoring test dependencies easily, my current idea is that LicenseFinder would treat each assembly as a dependency group. So If you're running LicenseFinder on a project like this example, and you wanted to ignore all libraries that only Api.Test depends on, you would run:

`license_finder ignored_groups add Api.Test`

That means every .net project using LicenseFinder would have different dependency group names, which is not what you see with Bundler or Npm. Still, I think it's a viable solution.

Thinking of submitting a pull request for this. Looking to get feedback on points 1&2 above.

mainej:
If you'd like to take this on, the [contributing guidelines](https://github.com/pivotal/LicenseFinder/blob/master/CONTRIBUTING.md) reference a [series of changes](https://github.com/pivotal/LicenseFinder/compare/v2.0.0...v2.0.1) which added rebar, the erlang package manager.  That diff is a good reference for everything that needs to be updated when adding a new package manager.

Regarding the licenseUrl, I think the best option is to write a `License.find_by_url`, similar to the existing `License.find_by_name`.  You can use it to override `Package#licenses_from_spec` in `NugetPackage`.  I can think of crazier solutions, but that seems like the best.

I like your idea for ignored groups.  Sounds like a good way to map the most similar concept.

LukeWinikates:
License.find_by_url is still in the works -- the naive implementation is painfully slow (because http) and has a low success rate (because most of the license pages are html and need to be parsed, which I didn't do in the first pass). I'll submit a PR if I come up with an implementation that I like.

PR #159, opened 2015-07-20
Add Nuget Support

LukeWinikates:
This PR adds basic support for the Nuget package manager as described in #155 . Dependency group support is in. License discovery will wait until later. For `LicenseFinder` will discover all the package dependencies for each assembly, and users will use `license_finder licenses add` to set the licenses manually.


mainej:
Looks like a good start.  Thanks for starting this.  Check out [how the rebar package manager was added](https://github.com/pivotal/LicenseFinder/compare/v2.0.0...v2.0.1).  If possible, please add feature specs and travis config, and update the CHANGELOG, README and Rakefile.  It would be nice if you mention the limitations around setting licenses in the README, at least until you implement the find-license-by-url idea.

LukeWinikates:
Thanks @mainej -- I've added the feature spec and updated documentation. Because there's no dependency on a command line tool, I don't think there's anything to add to the Rakefile or special travis config.

mainej:
:+1: Thanks!

PR #160, opened 2015-07-22
Use '&&' instead of ';' when chaining shell commands.

### LINKED ISSUE #158: Recent breakage of windows support
NPM and Bower support worked on windows until fairly recently. I think that the issue is the recent addition of `cd #{project_path};` to the NPM command and elsewhere. Windows' command line doesn't support `;`.

Could LicenseFinder use `&&` instead?

(+ @erd , whose name I recognize from the [relevant commit](https://github.com/pivotal/LicenseFinder/commit/97974291d741309d26298c6a2b0c3efc6d11b938))

erd:
Closes #158.

### DIFF
--- a/lib/license_finder/package_managers/bower.rb
+++ b/lib/license_finder/package_managers/bower.rb
@@ -11,7 +11,7 @@ module LicenseFinder
     private
 
     def bower_output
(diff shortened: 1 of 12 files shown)


LukeWinikates:
@erd It works! Thanks.

PR #169, opened 2015-08-24
Local config file

tinco:
Hi,

License finder would crash on our projects because they have a .bundle/config file that defined an alternative bundle_path. It took me many hours of looking through the bundler source code before I found out how I could get this to work.

As the code is now it will probably not work if you'd try to run it on multiple projects in the same process. I think I had that working in one of the between files but the required code was very ugly.

I spent too much time already for something that I should've done by hand in a few hours (indexing licenses of our projects) so I haven't written a spec for this. If you really want one before you merge the request I might find some time to do so, in the next weeks, but there's a good chance we'll have to pull in the super ugly hacky 'reset bundler state' code I reverted in this pr.

If you've got tips that'd be great as well ;)

### DIFF
--- a/lib/license_finder/package_managers/bundler.rb
+++ b/lib/license_finder/package_managers/bundler.rb
@@ -29,6 +29,7 @@ module LicenseFinder
 
         load 'bundler/rubygems_integration.rb'
         ENV['BUNDLE_GEMFILE'] = package_path.to_s
+        ::Bundler.setup
         @definition = ::Bundler::Definition.build(package_path, lockfile_path, nil)
 
         ENV['BUNDLE_GEMFILE'] = old_gemfile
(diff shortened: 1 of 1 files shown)


tinco:
Just to be clear: the PR as it is now fixes the problem for us.

mainej:
I'd like a spec, to make sure I understand the problem, to prevent regressions, for documentation/explanation, and to try to find a better solution.  Or, if you don't have time, could you provide the crash, your .bundle/config, and if possible, your Gemfile?

I'm hesitant to call `::Bundler.setup` unless we can't find any other way.  It changes the `$LOAD_PATH` for the license_finder code itself.

This is another scenario where it would be nice to switch away from using `::Bundler`, and instead shell out to `bundler list` or `gem list`, like we do with the other package managers.  However, neither of those commands can output all the data we need, and both seem to use a custom format (not JSON).

xtreme-lisheng-tai:
Closing this issue since it has been inactive for quite some time now.

PR #184, opened 2015-12-12
specified gem versions in requirements spec

humzashah:
the latest mime-type version was requiring a version of mime-types-data that required ruby version greater than 2.0. this is why mime-types version is now specified to one that can work with
older ruby versions.

similar ruby version problem was faced with the cocoapods gem on the darwin platform hence its version has been specified too.


flavorjones:
Thank you for this contribution! I had already fixed the mime-types dependency on master, but edited your commit during merging to keep the cocoapods version change.

Merged into master!

PR #185, opened 2015-12-23
Add support for approving particular versions of dependencies

### LINKED ISSUE #183: Approvals by dependency version
Working on a project with a client who does not approve Open Source projects on a per-license basis, but on a project-by-project basis, so I cannot whitelist a license and cal it done. Rather, I need to individually approve each dependency. Worse, they approve on a version-by-version basis for each dependency, so just because they've approved, say, `Junit 4.12` does *not* mean that `Junit 4.13` is approved, and I'd like a way to track that as well.

Currently, license_finder does not seem to provide a way to tell me that I've moved to a newer, unapproved, version of a dependency (or an older, unapproved version :-/).

I'd imagine this working something along the lines of `license_finder --approvals add <dependency>:<version>`. If you don't specify a version, the behavior is as-is (i.e. all versions are approved). Removals would be similar, with `approvals remove <dependency>` removing that dependency and all versions, and `approvals remove <dependency>:<version>` removing that specific version of that dependency.

ipsi:
Resolves #183 - you can now approve dependencies on a version-by-version basis. Might need more work to make it clear when a dependency is approved for a specific version, but you're using an earlier/later version, which is not approved.

### DIFF
--- a/lib/license_finder/cli/makes_decisions.rb
+++ b/lib/license_finder/cli/makes_decisions.rb
@@ -9,6 +9,7 @@ module LicenseFinder
         def auditable
           method_option :who, desc: "The person making this decision"
           method_option :why, desc: "The reason for making this decision"
+          method_option :version, desc: 'The version that will be approved'
         end
       end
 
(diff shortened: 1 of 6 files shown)


flavorjones:
This looks great! Thank you for the contribution, I've merged this into master and it will be available in v2.1.0.rc2 which I should cut today.

PR #186, opened 2015-12-30
Switch to using git submodules for go projects with .envrc file

jvshahid:
Previously `go list` was used to list the projects in the current
directory. This is unecessary since most projects that use .envrc
submodule their dependencies. Also, `go list` will report the project
in interest (as well as its dependencies) which is uneeded.

Signed-off-by: John Shahid <jshahid@pivotal.io>


flavorjones:
Merged manually. Thank you!

PR #210, opened 2016-03-04
Allow `groups` as an available column for CSV

jonwolski:
This change adds `groups` to the `AVAILABLE_COLUMNS` for `--format csv`.

Background
----------

It is useful to know the "groups" in which a dependency exists because
license restrictions may apply differently for different uses
(e.g. AGPL ramifications for SaaS on local development versus running on
a production server).


flavorjones:
This looks great, thank you for the contribution!

PR #233, opened 2016-07-04
Fix 'package_management_command()' to 'bundle'

raimon49:
Hi, I wanted to look for dependencies license with Bundler. However, I was not able to do it. At license_finder v2.1.2.

I tried `--debug` option and arrived at this correction.

### DIFF
--- a/lib/license_finder/package_managers/bundler.rb
+++ b/lib/license_finder/package_managers/bundler.rb
@@ -18,7 +18,7 @@ module LicenseFinder
     end
 
     def self.package_management_command
-      "bundler"
+      "bundle"
     end
 
     private


sschuberth:
Interesting. Usually this should not be reuiqred as a `bundler` script is installed as an almost exact copy of the `bundle` script:

    diff -u /usr/local/bin/bundle /usr/local/bin/bundler
    --- /usr/local/bin/bundle       2016-06-17 16:16:47.321186259 +0200
    +++ /usr/local/bin/bundler      2016-06-17 16:16:47.321186259 +0200
    @@ -20,4 +20,4 @@
     end

     gem 'bundler', version
    -load Gem.bin_path('bundler', 'bundle', version)
    +load Gem.bin_path('bundler', 'bundler', version)

But as I guess `bundle` is the more common way to execute Bundler, +1 from my side.

raimon49:
@sschuberth Thank for your commentary. Both commands surely existed in the local environment that I made using [rbenv](https://github.com/rbenv/rbenv).

As for the cause, there was only `bundle` command in my server environment.

And I agree when `bundle` command is more common way means.

flavorjones:
Thanks for submitting this! Merging.

raimon49:
@flavorjones Thanks!

PR #244, opened 2016-09-29
Enable '--columns' option with text reports

raimon49:
Hi, `--columns` option is very useful. I want to use it in `text` format, too.

It's keep compatibility by the default:

``` sh
$ bundle exec license_finder --format=text
...................................................................
sinatra, 1.0, MIT
```

When user appointed `--columns` option, `LicenseFinder` output it according to an appointed values:

``` sh
$ bundle exec license_finder --columns=name version homepage authors --format=text
...................................................................
sinatra, 1.0, http://sinatra.rubyforge.org, "Blake Mizerany, Ryan Tomayko, Simon Rozet"
```

### DIFF
--- a/lib/license_finder/reports/text_report.rb
+++ b/lib/license_finder/reports/text_report.rb
@@ -2,8 +2,11 @@ module LicenseFinder
   class TextReport < CsvReport
     COMMA_SEP =  ", "
 
-    def initialize(dependencies, options={})
-      super(dependencies, options.merge(columns: %w[name version licenses]))
+    def initialize(dependencies, options)
+      empty_options = {}
+      default_columns = %w[name version licenses]
+      columns = (Array(options[:columns]) & self.class::AVAILABLE_COLUMNS) || default_columns
+      super(dependencies, empty_options.merge(columns: columns))
     end
   end
 end
--- a/lib/license_finder/reports/text_report.rb
+++ b/lib/license_finder/reports/text_report.rb
@@ -5,7 +5,10 @@ module LicenseFinder
     def initialize(dependencies, options)
       empty_options = {}
       default_columns = %w[name version licenses]
-      columns = (Array(options[:columns]) & self.class::AVAILABLE_COLUMNS) || default_columns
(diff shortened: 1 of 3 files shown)


spec/lib/license_finder/reports/text_report_spec.rb
```diff
@@ -22,15 +22,20 @@ module LicenseFinder
         result
       end
 
-      subject { described_class.new([dep3, dep2, dep1]).to_s }
+      subject { described_class.new([dep3, dep2, dep1], {}).to_s }
```
raimon49: I'm sorry, I resolved that a test failed.

raimon49:
Hi, I rebase this branch to pass Travis CI.

I would like you to review this P-R in your backlog.

lib/license_finder/reports/text_report.rb
```diff
@@ -2,8 +2,14 @@ module LicenseFinder
...
-    def initialize(dependencies, options={})
-      super(dependencies, options.merge(columns: %w[name version licenses]))
+    def initialize(dependencies, options)
+      empty_options = {}
+      default_columns = %w[name version licenses]
+      columns = (Array(options[:columns]) & self.class::AVAILABLE_COLUMNS)
+      if columns.empty?
+        columns = default_columns
+      end
+      super(dependencies, empty_options.merge(columns: columns))
```
xtreme-jesse-malone: Curious about why the merge here is on `empty_options` and not `options`. It doesn't matter at the moment since `CsvReport` only pays attention to `:columns` but it seems possible to me that in the future `CsvReport` could use other options that would get squashed by replacing `options` with `empty_options`, as happens here. I could be missing something. Is there a reason for using `empty_options`?

raimon49: Thank you for the review.

As you say, using the `:columns` inherited from `CsvReport` is safer and simpler.

I tried correcting it. Can you review it again?

Manifaust:
@raimon49 thanks for your addition!

Issue #249, opened 2016-10-20
Drop support for Ruby 1.9.3 and Ruby 2.0


flavorjones:
Neither have upstream support any longer.

Discuss whether dropping these should be a major version bump or not.

connorshea:
If we're following SemVer, I'd say this should be a major version bump, but I don't feel too strongly about it.

Manifaust:
Removed with #279

PR #267, opened 2016-12-27
Add Ruby 2.4.0 to test matrix.

connorshea:
Happy holidays!


connorshea:
Nokogiri 1.7.0 was literally just released and it drops support for < Ruby 2.1. I guess we'll need to either use custom gemfiles for different Ruby versions or drop < Ruby 2.1.

connorshea:
@flavorjones well this is ironic, you're the one who released Nokogiri 1.7.0 :D 

How do you think we should resolve this?

flavorjones:
@connorshea can you help me understand what problem you're trying to solve?

flavorjones:
Also see possibly-related #249 which drops support for Ruby < 2.1.

These rubies have reached end-of-life and there's no need for us to continue to support them in LicenseFinder.

connorshea:
@flavorjones the problem is just that we need to either lock nokogiri or remove 2.1 support, that's what I was wondering. So I guess we should go with dropping 2.1?

flavorjones:
Sorry, feel the need to clarify: Ruby 2.1 is still supported; we're talking
about ending support for Ruby 1.9.3 and 2.0.x.

Another alternative is to release a version of LF that's pinned to Nokogiri
~> 1.6.8 before we make a release that removes support for older rubies.

Manifaust:
@connorshea thanks for your help!

PR #274, opened 2017-01-11
Update README.md

ddubson:
Specifying that, at this time, only CocoaPods 0.39 and below are supported. Linked to article describing what to do with CocoaPods 0.39 and below vs. 1.0 and above.


Manifaust:
Thanks @ddubson for clarifying update!

PR #278, opened 2017-01-17
Modifications for successfully running in ConcourseCI

ddubson:
The following changes to unit and feature tests were added to successfully run a multi Ruby pipeline in Concourse CI.

For commit 0dc3da4, we added the `--allow-root` flag to Bower since Concourse uses a Docker container to run its jobs, which runs processes as root user. Bower refuses to install under root by default.

For commit 9d3efc6, we modified `projects.rmtree` to `FileUtils.rmtree` due to a potential bug in JRuby 9.0.4.0 which refused to clear the temporary directory upon start of a test (after NPM feature test, and before Nuget feature test in particular).

### DIFF
--- a/features/support/testing_dsl.rb
+++ b/features/support/testing_dsl.rb
@@ -383,7 +383,8 @@ module LicenseFinder
 
       def reset_projects!
         # only destroyed when a test starts, so you can poke around after a failure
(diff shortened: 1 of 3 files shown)


Manifaust:
Changes will be included in #280

PR #281, opened 2017-02-06
support license_links csv option

lbalceda:
This change adds `license_links` to the AVAILABLE_COLUMNS for `--format csv`.

####  Background

Added in to have some parity with the `--report html` format (which has links to licenses) since I mainly use the csv format


xtreme-jesse-malone:
Thanks for this!

Issue #218, opened 2016-05-02
Aliases


connorshea:
Is it possible to add an alias to, e.g. some gems are "LGPL-2.1+" and I'd like to add that as an alias for LGPL. I can't find any other issues about this, and it doesn't seem to be documented anywhere.

Thanks, Connor

wayne-luminal:
+1 This would be nice

flavorjones:
Hi, @connorshea.

Thanks for opening this issue. I'm not sure I totally understand, can you be more specific by what you mean by 'alias'? Are you asking for the emitted license name to be different in reports? Or are you asking for different whitelist behavior?

connorshea:
@flavorjones essentially, instead of having to whitelist "LGPLv2", "LGPL version 2", "LGPLv2+", "Lesser General Public License version 2", etc. separately I'd like to be able to whitelist multiple at once using "aliases". 

Instead of
```
- - :whitelist
  - LGPLv2
  - :who: Connor Shea
    :why: http://www.gnu.org/licenses/license-list.html#LGPLv2.1
    :versions: []
    :when: 2016-05-02 05:32:48.645841000 Z
- - :whitelist
  - LGPLv2+
  - :who: Connor Shea
    :why: http://www.gnu.org/licenses/license-list.html#LGPLv2.1
    :versions: []
    :when: 2016-05-02 05:32:48.645841000 Z
- - :whitelist
  - LGPL version 2
  - :who: Connor Shea
    :why: http://www.gnu.org/licenses/license-list.html#LGPLv2.1
    :versions: []
    :when: 2016-05-02 05:32:48.645841000 Z
```

It could be

```
- - :whitelist
  - LGPLv2, LGPLv2+, LGPL version 2, Lesser General Public License version 2
  - :who: Connor Shea
    :why: http://www.gnu.org/licenses/license-list.html#LGPLv2.1
    :versions: []
    :when: 2016-05-02 05:32:48.645841000 Z
```

Essentially the problem is that right now in GitLab we've had to make [repetitive additions to our decisions YML file](https://gitlab.com/gitlab-org/gitlab-ce/commit/5bb3e1934c2a059c8e7cdc4ecbe597781f502a3e) because they spelled the license ever so differently even though we already approved of the license in question.

Other than that license_finder has been great! Thank you so so much for the gem, I was actually putting together a *manual list* of all the gems we were using (hundreds) that didn't have certain licenses before I found this gem. I felt like a real dunce :D

[...]

PR #289, opened 2017-03-20
Adds `spdx` field for licenses

petems:
* Initial work to get spdx matching for licenses

I want to see if I'm on the right track for this

Related https://github.com/pivotal/LicenseFinder/issues/218

### DIFF
--- a/lib/license_finder/license.rb
+++ b/lib/license_finder/license.rb
@@ -1,3 +1,5 @@
+require 'spdx-licenses'
+
 require "license_finder/license/text"
 require "license_finder/license/template"
 
@@ -23,6 +25,10 @@ module LicenseFinder
       def find_by_text(text)
         all.detect { |l| l.matches_text? text }
       end
+
+      def find_by_spdx(spdx_string)
+        all.detect { |l| l.spdx == spdx_string && l.valid_spdx?(l.spdx) }
+      end
     end
 
     def initialize(settings)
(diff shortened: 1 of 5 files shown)


kdykeman:
Thanks @petems. From what I'm seeing so far it looks to be on the right track. I'll get another set of eyes to look the PR over within the next week.

I'm not sure if you were already thinking this as well, but it would probably also be friendlier to downstream tools if the reports could optionally list licenses using their spdx names.  That could be a separate PR.

Manifaust:
@petems thanks for the PR, we have a couple questions.

You implemented `spdx` as an array, but aren't spdx entries unique, that is, there is a single name/id for each license in the [official list of spdx licenses](https://spdx.org/licenses/)?

Second, since there are so many licenses in that list, is it a good idea to introduce a new hardcoded property in the list of licenses in `Definitions`? I'm not sure yet what is the best way to go forward. But @flavorjones mentioned in the tagged issue the gem [spdx-licenses](https://github.com/domcleal/spdx-licenses). Maybe instead of introducing a new field, we can create a function that delegates to that gem? What do you think?

petems:
@Manifaust The gem is just for checking that a license given is actually a valid SPDX code, you can't use it for detecting. So I think we stilll need the field.

But it's a good idea to add it in for validation, so I did! 😄 

@kdykeman Can I get some more eyes on this PR? 😄

xtreme-lisheng-tai:
closing, since this has been opened for quite awhile.

Issue #290, opened 2017-03-21
LF doesn't recognize .NET projects using 'packages' directory instead of '.nuget' for dependencies


bspeck:
As a user working on the standard Visual Studio ASP.NET sample I expect License Finder to scan my dependencies listed in the `packages` folder where visual studio auto-populates them.

Currently I will see the following output:
```
.....................
All dependencies are approved for use
```

Which can be taken as a false positive.

bspeck:
Submitted pull request #292 to address this issue

kdykeman:
Thanks for the PR! I've merged it in.

PR #292, opened 2017-03-21
Fix for issue #290 (LF doesn't recognize .NET projects using 'packages' directory instead of '.nuget' for dependencies)

bspeck:
Add check for whether .nuget directory is present, if it isn't look for a packages directory.


bspeck:
@kdykeman @flavorjones This is a PR for the stuff we talked about on Slack.
Thanks!

Issue #235, opened 2016-07-11
is the first time user experience too mysterious?


LukeWinikates:
I asked my team to add license_finder to a gradle project and it was a rocky start. Turns out the problem was mostly #222. With the gradle discovery broken, the feedback from license_finder was misleadingly happy -- a few dots then exit `0`. If they were adding license_finder to the job without somebody around (me) who knows how the gem normally behaves, there wouldn't have been any way for them to know something was wrong.

I wonder if the default output should be more verbose, to make it more obvious when something is totally broken, and to set user expectations a bit better.

```
discovering package managers...
found gradle
found npm
discovering dependencies...
found 0 dependencies
```

flavorjones:
@LukeWinikates this feels like a vote to fix #222, would you mind if I closed this?

LukeWinikates:
It's not a duplicate of #222. I'm saying that the behavior of having effectively zero output if everything *seems* to go well makes the gem hard to use, and makes it hard to tell if something surprising has happened (like gradle not being discovered).

IOW, #222 is obviously a bug, but is it also a usability bug that #222 is effectively undiscoverable? Should the tool give more feedback about what it's doing, so that debugging is easier?

Package manager discovery is kind of magical but there's not a good way to sniff out whether the discovery is working or not.

mainej:
Looks like it would be pretty easy to have
`LicenseFinder::CLI::Main#action_items` print a message when
`license_finder.acknowledged.empty?` It could also return a non-zero exit
code, though that may be incorrect because it really is possible to have no
dependencies. Maybe a PR would clarify the scope and utility of the change.

flavorjones:
There's a `--debug` option, but it's incompletely and inconsistently implemented. In your case, @LukeWinikates, it should tell you that gradle isn't present.

I agree that output from LF is terrible, but unfortunately we have a very limited amount of engineering time that we can put towards overhauling UX. I'll leave this open.

LukeWinikates:
I wasn't aware of the debug option, and I am happy to know it exists! Although Nuget is missing from the list it prints out... 🤔 

I'm interested in doing a PR. Honestly, I think my first suggestion would be to show the debug output, and possibly a count of the discovered dependencies per package manager. That would make the current output more verbose, but would make it more obvious what's happening under the hood.

[...]

PR #293, opened 2017-03-23
Partial(?) fix for #235

bspeck:
Print a message when LF can't find any packages.

This isn't the most comprehensive fix for #235 but at least gets rid of the misleading false positive case.

Sample output:
```
pivotal@DESKTOP-KJBQ0TC:~/temp$ license_finder
........................
No dependencies recognized!
```

Thanks, @flavorjones @kdykeman @LukeWinikates !

### DIFF
--- a/features/features/cli/cli_spec.rb
+++ b/features/features/cli/cli_spec.rb
@@ -36,7 +36,7 @@ describe "License Finder command line executable" do
 
     developer.run_license_finder
     expect(developer).to be_receiving_exit_code(0)
-    expect(developer).to be_seeing 'All dependencies are approved for use'
+    expect(developer).to be_seeing 'No dependencies recognized!'
   end
 
   specify "displays an error if project_path does not exist" do
--- a/lib/license_finder/cli/main.rb
+++ b/lib/license_finder/cli/main.rb
@@ -34,11 +34,17 @@ module LicenseFinder
       desc "action_items", "List unapproved dependencies (the default action for `license_finder`)"
 
       def action_items
+        any_packages = license_finder.any_packages?
         unapproved = license_finder.unapproved
         blacklisted = license_finder.blacklisted
 
         # Ensure to start output on a new line even with dot progress indicators.
         say "\n"
+
+        unless any_packages
+          say "No dependencies recognized!", :red
+          exit 1
+        end
 
         if unapproved.empty?
           say "All dependencies are approved for use", :green
--- a/lib/license_finder/cli/main.rb
+++ b/lib/license_finder/cli/main.rb
@@ -43,7 +43,7 @@ module LicenseFinder
 
         unless any_packages
           say "No dependencies recognized!", :red
-          exit 1
+          exit 0
         end
 
         if unapproved.empty?
--- a/lib/license_finder/core.rb
+++ b/lib/license_finder/core.rb
@@ -37,7 +37,7 @@ module LicenseFinder
     end
 
     extend Forwardable
-    def_delegators :decision_applier, :acknowledged, :unapproved, :blacklisted
+    def_delegators :decision_applier, :acknowledged, :unapproved, :blacklisted, :any_packages?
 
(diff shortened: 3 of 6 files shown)


LukeWinikates:
I haven't tried this out locally, but I like the idea of getting something out there for this case.

I'm not sure if anyone ever intentionally runs license finder on a project that has no dependencies. Hard to imagine anything that would be its own standalone project with absolutely zero dependencies. If we were worried about the `exit 1` frustrating users, would we want a flag to opt-out? `--suppress-failure-on-zero-discovered-dependencies`

kdykeman:
Issue #235 is primarily about the confusing circumstance of getting no output from LicenseFinder to indicate that something may have gone wrong in certain circumstances - such as when configuring it for a new project. The "No dependencies recognized!" message seems to address the crux of Issue #235.

The `exit 1` breaks existing behavior where LicenseFinder is actually running without issue on a zero-dependency project. It also seems a bit adjacent to the "no output" problem. Full disclosure, the team I work with has a few projects that legitimately have no dependencies and a non-zero exit code would turn pipelines red.

We can give more consideration to using a flag to influence the exit code behavior if somebody files a separate issue describing their use case for wanting a non-zero exit.

LukeWinikates:
@kdykeman I suggested the non-zero exit code in this comment: https://github.com/pivotal/LicenseFinder/issues/235#issuecomment-270205263

I still think `exit 1` is the right behavior in cases when LicenseFinder *should* know that the user is trying to use it with a specific package manager, but fails to detect that package manager because of something subtly amiss -- like LF being picky about how you're using NuGet, or because you're using Gradle but don't have the other extra things right (https://github.com/pivotal/LicenseFinder#gradle-projects). I think I've had issues with npm output in older npm versions not working with license finder and causing it to not detect any dependencies, but now we're in ancient anecdotal evidence territory.

I'm on board with removing the `exit 1` here. I think the warning is a step in the right direction.

@bspeck @alidaka would it be possible to print a message for something like "0 package managers activated"? Or maybe to always print the number of package managers activated? 

I think it's really not the 0 dependencies that's the problem, it's the "we didn't find any package managers" bit that I think should be a red flag that the users see. I don't remember if debug mode already does this, but I think it'd be reasonable for it to be printed by default.

bspeck:
@LukeWinikates an output message "we didn't find any package managers" would be misleading in this case. This change doesn't check if package managers were found, it checks if any packages were found. Packages in this case map to dependencies, not package managers. 

What you are suggesting would be a different change. I think it's definitely a valuable message to print. When @alidaka and I were working on this change we determined that it'd require some refactoring for that information to be bubbled up from package_manager's current_packages method to main.rb where the message logging logic is.

We started playing with license finder to figure out how it worked for ASP.NET apps. It turned out that it didn't work for ASP.NET apps with dependencies listed in the packages directory. It took us a while to realize this was the case when the output said "All dependencies are approved for use". The "No dependencies recognized!" output would have helped in our case.

bspeck:
@LukeWinikates and @kdykeman, @emgord and I updated the change to return 0. We think it would be appropriate to return an error code at a later release along with an opt out flag, but want users to have a chance to adjust.

kdykeman:
It looks like one of the tests is failing now in Concourse for the PR - https://osl.ci.cf-app.com/teams/main/pipelines/LicenseFinder/jobs/PR-ruby-2.4.0/builds/21

LukeWinikates:
@bspec - sounds like "no package managers found" is a big enough effort to warrant its own pull request. Thanks for the context. This PR makes me happy.

bspeck:
@kdykeman - we updated that test. We weren't able to reproduce the failure locally, because we haven't been able to run tests against the docker image. We're working on a dual boot windows/mac machine, and docker requires some changes to the Windows Bios setting that are tricky to get to through bootcamp. We hope this addresses the failure in as it is exhibited in the CI environment.

PR #312, opened 2017-07-12
Update Ruby dependency. Closes #311.

### LINKED ISSUE #311: LicenseFinder 3.x fails for Ruby < 2.3
# Summary

LicenseFinder 3.x fails for Ruby 1.9.3: https://travis-ci.org/cucumber/aruba/jobs/252903532#L348. You may want to adjust the Ruby version in your [`gemspec`](https://github.com/pivotal/LicenseFinder/blob/master/license_finder.gemspec#L8).

maxmeyer:
This  should fix #311 where the "Safe Navigation Operator" makes a build fail.

### DIFF
--- a/license_finder.gemspec
+++ b/license_finder.gemspec
@@ -5,7 +5,7 @@ require 'license_finder/platform'
 require 'license_finder/version'
 
 Gem::Specification.new do |s|
-  s.required_ruby_version = '>= 1.9.3'
+  s.required_ruby_version = '>= 2.3'
   s.name        = "license_finder"
   s.version     = LicenseFinder::VERSION
 


Nitrodist:
Shouldn't CI be running against this? Weird.

maxmeyer:
Should be, but nonetheless it not be possible to install the gem for rubies < 2.3.

kdykeman:
Holding this one for now - I think there are some things we can do to go back to supporting older versions.

kdykeman:
I also opened #321 requesting that Concourse CI status badges be added to PRs.

maxmeyer:
@kdykeman Thanks.

kdykeman:
Support for older versions of Ruby should be back as of 3.0.2.
CI status badges on PRs should also be working at this point.

Issue #314, opened 2017-07-13
LicenseFinder 3.0.1 incorrectly parsing some licenses


richardowen:
Create a package.json with this in it:

```json
{
  "dependencies": {
    "array-unique": "0.2.1"
  }
}
```

Running with LicenseFinder 3.0.0 you get the following:

```
$ license_finder version
3.0.0
$ license_finder
...................
Dependencies that need approval:
array-unique, 0.2.1, MIT
```

Running with LicenseFinder 3.0.1 you get the following:

```
$ license_finder version
3.0.1
$ license_finder
...................
Dependencies that need approval:
array-unique, 0.2.1, "{""type""=>""MIT"", ""url""=>""https://github.com/jonschlinkert/array-unique/blob/master/LICENSE""}"
```

kdykeman:
Fixed in commit 0e6d67c

PR #319, opened 2017-07-17
Add support for idiosyncratic (outdated) package.json format

### LINKED ISSUE #317: Invalid value for NPM "licenses" key causes an error
When an NPM package incorrectly provides a string for the "licenses" key, LicenseFinder raises an error.

See: flesler/jquery.localScroll#36

This is similar to #109, but for a String value rather than a Hash/Object. Incidentally, I don't think the fix applied for #109 works in the latest version of LicenseFinder. It looks like the spec for that scenario may have been lost during a rewrite.

A possible fix in [npm.rb](https://github.com/pivotal/LicenseFinder/blob/master/lib/license_finder/package_managers/npm.rb#L62) might look something like this:

```
    def licenses(dep_matches)
      licenses_lists = dep_matches.map do |match|
        match['licenses']
      end

      licenses_lists.reject! {|licenses| licenses.nil? || licenses == '[Circular]'}
      licenses_lists = licenses_lists.uniq
      case licenses_lists.count
        when 0
          dep_matches.map {|dep_match| dep_match['license']}.reject(&:nil?).uniq
        when 1
          normalize_license_list(licenses_lists.first)
        else
          raise "Varying lists of licenses provided for #{dep_matches.first['name']} (#{dep_matches.first['version']})"
      end
    end

    def normalize_license_list(license_list)
      case license_list
      when Hash
        [license_list['type']]
      when String
        [license_list]
      when Array
        license_list.map {|license| license['type']}
      when Nil
        []
      else
        raise "Invalid license: #{license_list}"
      end
    end
```

ryanmattcollins:
Resolves #317.

[Finish #148924915]

Signed-off-by: Jamie Klassen <cklassen@pivotal.io>

### DIFF
--- a/lib/license_finder/package_managers/npm.rb
+++ b/lib/license_finder/package_managers/npm.rb
@@ -1,96 +1,52 @@
-require 'yajl'
+require 'json'
 require 'tempfile'
 
 module LicenseFinder
   class NPM < PackageManager
-    DEPENDENCY_GROUPS = ["dependencies", "devDependencies"]
-
     def current_packages
-      top_level_deps = npm_json['dependencies']&.values || []
-      package_json = JSON.parse(File.read(package_path), :max_nesting => false)
-      top_level_deps.each do |dep|
-        dep['groups'] = DEPENDENCY_GROUPS.select do |group|
-          package_json[group]&.keys&.include? dep['name']
-        end
-      end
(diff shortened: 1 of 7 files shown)


kdykeman:
Also resolved #314

PR #354, opened 2017-09-21
use minimum JIT for jruby to improve test time

pivotal-jamie-klassen:
- following advice from
  - http://blog.headius.com/2010/03/jruby-startup-time-tips.html
  - https://github.com/jruby/jruby/wiki/Improving-startup-time

[finish #151267640]

Signed-off-by: Yash Pancholi <ypancholi@pivotal.io>


pivotal-jamie-klassen:
Abandoning this... the time invested is yielding diminishing returns.

PR #382, opened 2017-10-24
Add support for prepare step

xtreme-lisheng-tai:
# Prepare
### Technical notes
- Initial work for `--prepare` step
- `action_items` and `report` tasks now support a `--prepare` switch
- Passing in the `--prepare` switch will result in `prepare` method being called on each package manager which is active in the current project path. 
- Currently none of the package managers fully implement the prepare step. (experimental GVT support is currently included though)

### DIFF
--- a/lib/license_finder/cli/main.rb
+++ b/lib/license_finder/cli/main.rb
@@ -39,7 +39,7 @@ module LicenseFinder
         method_option :debug,
                       :aliases => '-d',
                       :type => :boolean,
-                      :desc => 'emit detailed info about what LicenseFinder is doing'
+                      :desc => 'Emit detailed info about what LicenseFinder is doing'
 
         method_option :prepare,
                       :aliases => '-p',
--- a/lib/license_finder/cli/main.rb
+++ b/lib/license_finder/cli/main.rb
@@ -17,27 +17,45 @@ module LicenseFinder
         'csv' => CsvReport
       }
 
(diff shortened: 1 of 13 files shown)


lib/license_finder/cli/main.rb
```diff
@@ -17,27 +17,45 @@ class Main < Base
...
+      class_option :subprojects, type: :array, desc: %q(Generate a single report for multiple sub-projects. Ex: --subprojects='path/to/project1', 'path/to/project2')
+      class_option :recursive, desc: 'Recursively runs License Finder on all sub-projects.'
+
+      # Method options which are shared between report and action_item
+      def self.shared_options
+
+        method_option :debug,
+                      :aliases => '-d',
+                      :type => :boolean,
+                      :desc => 'emit detailed info about what LicenseFinder is doing'
```
ryanmattcollins: emit should be capitalized

lib/license_finder/package_manager.rb
```diff
@@ -10,44 +10,51 @@ module LicenseFinder
...
-      logger.installed self, false
-      return false
-    end
+      def package_management_command
+        nil
+      end
 
-    def self.package_management_command
-      nil
+      def prepare_command
```
ryanmattcollins: Might be nice to have comments indicating which methods of PackageManager must/should be included in subclasses

lib/license_finder/cli/main.rb
xtreme-andy-shen: Will do.

PR #385, opened 2017-10-26
Refactor Logger

xtreme-andy-shen:
Decentralize use of Logger.
Use Logger.green to produce green string; Logger.red to produce red string
To log information, use logger.log and pass in the class and the string

[#152282179]

Signed-off-by: Vikram Yadav <vyadav@pivotal.io>

### DIFF
--- a/lib/license_finder/package_manager.rb
+++ b/lib/license_finder/package_manager.rb
@@ -36,17 +36,15 @@ module LicenseFinder
 
       def installed?(logger=Core.default_logger)
         if package_management_command.nil?
-          logger.installed self, "no command defined" # TODO comment me out
+          logger.log self, 'no command defined' #TODO comment me out
           return true
-        end
-
-        if command_exists?(package_management_command)
-          logger.installed self, true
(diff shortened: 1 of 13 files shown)


lib/license_finder/package_manager.rb
```diff
@@ -69,7 +67,10 @@ def active?
       path = detected_package_path
       self.class.installed?(logger) &&
           !path.nil? &&
-        path.exist?.tap { |is_active| logger.active self.class, is_active }
+        path.exist?.tap do |is_active|
+          message = is_active ? Logger.green("is active") : 'is not active'
+          logger.log self.class, message
```
ryanmattcollins: Just double checking:
Seems like there is a Class method on `Logger` for `.green` and `.red`, but then in this method `logger` is an instance being passed in and `#log` is being called on it. Is that intentional?

My first impression implies that `Logger.green("Hello World")` would print out "Hello World" in green. Instead, it seems like it's colour coding the message. I feel like having that be a part of the `.log` signature rather than encoding the string with it is more intuitive.

PR #440, opened 2017-12-12
Refactor cli layer

xtreme-lisheng-tai:
## Configuration and Options refactor
* Refactor and clean up CLI layer
* Logger now no longer takes a hash consisting of quiet / debug but rather just a mode
[CHANGED] Class options format and columns are now method options specific to report since they have no effect on either action_items nor diff 
* core no longer holds onto the decisions object thus forcing main to keep a random core around even when it isn't being used
* Spec file for main now mimics Thor situations more accurately

### DIFF
--- a/spec/lib/license_finder/cli/main_spec.rb
+++ b/spec/lib/license_finder/cli/main_spec.rb
@@ -85,11 +85,13 @@ module LicenseFinder
         end
 
         it 'passes the config options to the new LicenseFinder::LicenseAggregator instance' do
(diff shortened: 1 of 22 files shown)


spec/lib/license_finder/cli/main_spec.rb
```diff
@@ -104,14 +107,13 @@ def report
           expect(report).to eq "\"one dependency\", 1.1, unknown\n"
         end
 
-        it 'will output a specific format' do
-          subject.options = { format: 'markdown' }
-
+        it 'will output a specific format', :focus do
```
xtreme-andy-shen: Please leave focus

PR #449, opened 2018-01-11
[Added] Specify log directory for prepare

pivotal-dk:
Signed-off-by: Ryan Collins <rcollins@pivotal.io>

### DIFF
--- a/lib/license_finder/core.rb
+++ b/lib/license_finder/core.rb
@@ -55,6 +55,7 @@ module LicenseFinder
     end
 
     def prepare_projects
+      clear_logs
       package_managers = @scanner.active_package_managers
       package_managers.each do |manager|
         logger.debug manager.class, 'Running prepare on project'
@@ -80,10 +81,15 @@ module LicenseFinder
       @scanner.active_packages
     end
 
+    def clear_logs
+      FileUtils.rm config.log_directory if File.directory? config.log_directory
+    end
+
     def options
       {
         logger: logger,
         project_path: config.project_path,
+        log_directory: File.join(config.log_directory, project_name),
         ignored_groups: decisions.ignored_groups,
         go_full_version: config.go_full_version,
         gradle_command: config.gradle_command,
--- a/lib/license_finder/core.rb
+++ b/lib/license_finder/core.rb
@@ -82,7 +82,7 @@ module LicenseFinder
     end
 
     def clear_logs
-      FileUtils.rm config.log_directory if File.directory? config.log_directory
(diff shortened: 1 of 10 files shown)


lib/license_finder/core.rb
```diff
@@ -80,10 +81,15 @@ def current_packages
...
 
+    def clear_logs
+      FileUtils.rm config.log_directory if File.directory? config.log_directory
+    end
+
     def options
       {
         logger: logger,
         project_path: config.project_path,
+        log_directory: config.log_directory,
```
xtreme-vikram-yadav: can we set this directly to what is merged on line 34?

lib/license_finder/package_manager.rb
```diff
@@ -111,11 +112,9 @@ def log_errors(stderr)
     end
 
     def log_to_file(contents)
-      log_dir = File.join(@project_path, 'lf_logs')
-      log_file_path = File.join(log_dir, 'error.log')
-
-      FileUtils.mkdir_p(log_dir)
-      File.open(log_file_path, 'w') do |f|
+      FileUtils.mkdir_p @log_directory
+      log_file = File.join(@log_directory, "#{self.class.package_management_command || 'errors'}.log")
```
xtreme-vikram-yadav: Just a minor thing and It may be very unlikely but some package managers might create log files based on their names

spec/lib/license_finder/package_manager_spec.rb
```diff
@@ -73,59 +73,59 @@ module LicenseFinder
 
     describe '#prepare' do
       context 'when there is a prepare_command' do
+        let(:prepare_error) { /Prepare command .* failed/ }
+        let(:project_path) { '/path/to/project' }
+        let(:log_directory) { '/path/to/project/logs/project' }
+        let(:log_file_path) { File.join(log_directory, 'errors.log') }
```
xtreme-vikram-yadav: we should add test to cover existence of .package_management_command

lib/license_finder/core.rb
pivotal-dk: Yep, sounds good.

lib/license_finder/package_manager.rb
pivotal-dk: We can prepend log files with 'prepare_' to avoid confusion

spec/lib/license_finder/package_manager_spec.rb
pivotal-dk: Sure thing.

xtreme-vikram-yadav:
👍

PR #457, opened 2018-01-30
[FIXED] Fix yarn not working with --project_path option

xtreme-lisheng-tai:
[finish #154696328]

Simple fix for yarn 

Signed-off-by: Li Tai <ltai@pivotal.io>

### DIFF
--- a/lib/license_finder/package_managers/yarn.rb
+++ b/lib/license_finder/package_managers/yarn.rb
@@ -7,10 +7,9 @@ module LicenseFinder
     end
 
     def current_packages
-
       cmd = Yarn::SHELL_COMMAND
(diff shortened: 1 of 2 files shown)


lib/license_finder/package_managers/yarn.rb
```diff
@@ -7,7 +7,7 @@ def possible_package_paths
     end
 
     def current_packages
-      stdout, _stderr, status = Cmd.run(Yarn::SHELL_COMMAND)
+      stdout, _stderr, status = Cmd.run(Yarn::SHELL_COMMAND + " --cwd #{project_path}")
```
ryanmattcollins: What if project path is nil?

PR #494, opened 2018-05-25
Upgrade Dockerfile base to xenial

Manifaust:


xtreme-lisheng-tai:
probably needs rebase and seems to cause failures in CI test

xtreme-shane-lattanzio:
Looks like tests may have a false positive. It doesn't look like they actually ran and there is an error of some kind in the log.

xtreme-shane-lattanzio:
PR is not updating. Recreating

PR #495, opened 2018-05-31
Add project paths command to LF

xtreme-shane-lattanzio:
Signed-off-by: Daniil Kouznetsov <dkouznetsov@pivotal.io>

### DIFF
--- a/features/features/cli/cli_spec.rb
+++ b/features/features/cli/cli_spec.rb
@@ -120,7 +120,7 @@ describe 'License Finder command line executable' do
 
       specify 'returns current path' do
         developer.execute_command(license_finder_command)
-        expect(developer).to be_seeing_something_like %r{^#{Regexp.escape(@project.project_dir.to_s)}$}
+        expect(developer).to be_seeing_something_like /^#{Regexp.escape(@project.project_dir.to_s)}$/
       end
     end
   end
--- a/features/features/cli/cli_spec.rb
+++ b/features/features/cli/cli_spec.rb
@@ -99,7 +99,7 @@ describe 'License Finder command line executable' do
     end
   end
 
-  context 'running project_roots', :focus do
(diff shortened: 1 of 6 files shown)


lib/license_finder/project_finder.rb
```diff
@@ -41,7 +42,7 @@ def project_root?(pathname)
 
     def active_project?(project_path)
       active_project = @package_managers.map do |pm|
-        pm.new(project_path: project_path).active?
+        pm.new(project_path: project_path).active? unless pm == GoWorkspace && @configurable
```
mainej: Why the break in polymorphism? That is, why not pass `@configurable` to `pm.new`, and let GoWorkspace do something with it?

A related problem is that the name `@configurable` doesn't imply what it will be used for. Instead, maybe you could add an argument (possibly exposed at the CLI) which says which types of project managers to skip.

features/features/cli/cli_spec.rb
```diff
@@ -98,4 +98,30 @@
       expect(developer).to_not be_seeing('MIT')
     end
   end
+
+  context 'running project_roots', :focus do
```
xtreme-vikram-yadav: :focus .

xtreme-vikram-yadav: can you update this to a describe?

lib/license_finder/cli/main.rb
```diff
@@ -158,11 +164,12 @@ def check_valid_project_path
         raise "Project path '#{config.project_path}' does not exist!" unless config.valid_project_path?
       end
 
-      def aggregate_paths
+      def aggregate_paths(strict_matching = false)
```
xtreme-vikram-yadav: Can we use the config to do this instead? That way we avoid having a parameter on this method just to pass it to the project finder.

PR #498, opened 2018-06-08
Follow PyPI redirects. Fixes #497

### LINKED ISSUE #497: LicenseFinder v5.3.0 doesn't follow PyPI redirects
The change from httparty to Net::HTTP means that redirects are no longer followed automatically. For example, when looking up the licence for cycler, LicenseFinder makes a request to https://pypi.org/pypi/cycler/0.10.0/json. This results in a redirect to https://pypi.org/pypi/Cycler/0.10.0/json (note the capital C). Previously httparty would have followed this redirect and found the licence information. However Net::HTTP does not do this and so, since the response code is a 301 (and not a 200), the licence information is not returned.

The user-facing result is:

```
$ license_finder version
5.2.3
$ license_finder
LicenseFinder::Pip: is active

cycler, 0.10.0, BSD
six, 1.11.0, MIT
$ gem install license_finder
Fetching: license_finder-5.3.0.gem (100%)
Successfully installed license_finder-5.3.0
1 gem installed
$ license_finder version
5.3.0
$ license_finder
LicenseFinder::Pip: is active

cycler, 0.10.0, unknown
six, 1.11.0, MIT
```

On version 5.2.3, the licence for cycler is correctly detected as BSD. But on 5.3.0 the licence is unknown.

richardowen:


xtreme-shane-lattanzio:
@richardowen Thanks for the PR! Looks good but would you mind writing a quick test for this in the pip_spec? We already have some stubbed responses going on in there so you would just need to follow suit.

richardowen:
@xtreme-shane-lattanzio Apologies for the delay but have added a test for this now. Verified that the test fails before my change and passes after.

xtreme-shane-lattanzio:
@richardowen Thanks. Looks good!

PR #499, opened 2018-06-09
Add nuget prepare

### LINKED ISSUE #496: Add nuget prepare command
I would be happy to add package restore for NuGet.

For the Dockerfile that would be adding mono-complete & nuget. This would be adding

```Dockerfile
# TODO: add nuget
# https://docs.microsoft.com/en-us/nuget/install-nuget-client-tools#macoslinux
RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF &&\
  #apt-get install apt-transport-https &&\
  echo "deb https://download.mono-project.com/repo/ubuntu stable-trusty main" | sudo tee /etc/apt/sources.list.d/mono-official-stable.list &&\
  apt-get update &&\ 
  apt-get install -y mono-complete &&\
  curl -o /usr/local/bin/nuget.exe https://dist.nuget.org/win-x86-commandline/latest/nuget.exe &&\
  echo "alias nuget=\"mono /usr/local/bin/nuget.exe\"" >> ~/.bash_aliases
```

For the `nuget.rb` that would be adding

```ruby
def self.prepare_command
   'nuget restore'
end
```

mkoertgen:
Resolves #496

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -119,6 +119,15 @@ RUN apt-get install -y python-dev && \
 # install Cargo
 RUN curl -sSf https://static.rust-lang.org/rustup.sh | sh -s -- --disable-sudo
 
+# install NuGet (w. mono)
+# https://docs.microsoft.com/en-us/nuget/install-nuget-client-tools#macoslinux
+RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF &&\
+  echo "deb https://download.mono-project.com/repo/ubuntu stable-trusty main" | sudo tee /etc/apt/sources.list.d/mono-official-stable.list &&\
+  apt-get update &&\ 
+  apt-get install -y mono-complete &&\
+  curl -o /usr/local/bin/nuget.exe https://dist.nuget.org/win-x86-commandline/latest/nuget.exe &&\
+  echo "alias nuget=\"mono /usr/local/bin/nuget.exe\"" >> ~/.bash_aliases
+
 # install license_finder
 COPY . /LicenseFinder
 RUN bash -lc "cd /LicenseFinder && bundle install -j4 && rake install"
--- a/lib/license_finder/package_managers/nuget.rb
+++ b/lib/license_finder/package_managers/nuget.rb
@@ -26,8 +26,12 @@ module LicenseFinder
       path = project_path.join('vendor/*.nupkg')
       nuget_dir = Dir[path].map { |pkg| File.dirname(pkg) }.uniq
 
+      path = project_path.join('*.sln')
+      solution_file = Dir[path].first
+
       possible_paths = [project_path.join('packages.config'), project_path.join('.nuget')]
       possible_paths.unshift(Pathname(nuget_dir.first)) unless nuget_dir.empty?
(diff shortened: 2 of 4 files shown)


xtreme-shane-lattanzio:
After rerunning. It also looks like the tests are failing for reasons in this PR as well.

pivotal-dk:
Hey there, @mkoertgen ! When you ran the tests locally, did you run them from the docker image you built using the changes in your `Dockerfile`?. Based on a quick look through our CI it looks like your commands are being run through the docker build but this would also fail when running locally. Could you verify this? Thanks for the PR!

pivotal-dk:
Adding on to the previous comment, it seems that running `rake spec` and `rake features` error with `LicenseFinder::Nuget: is not installed` after building your docker image. Please ensure these two pass. The reason it's failing is because `rake spec` and `rake features` are better end-to-end tests; they initialize a new `Scanner` which checks if Nuget is installed.

mkoertgen:
No, just ran the test locally (not inside docker). And only the nuget tests.

Ok, will build & re-run using docker then.

mkoertgen:
Looks good now

xtreme-shane-lattanzio:
@mkoertgen Sorry for the delayed response. Thanks for looking into fixing the docker tests. I don't think it is a good idea to remove tests to get them to pass but I understand where you are coming from. A better approach is probably to override `Nuget.installed?` and in turn the `command_exists?` functionality to use `type` instead of `which` for the nuget package manager. After that is done and the other tests are added back, we should be good to merge. Thanks again!

mkoertgen:
Ok, good point. I agree that's the better way to go.

mkoertgen:
Sadly, i found no good way to have `Cmd.run -> Open3.capture3` act on shell built-ins like `type`.
So i ended up overriding all platform-specific commands and DRYed up tests using _shared_examples_, then stubbing the platform in different _contexts_ for _windows_ and _linux_

mkoertgen:
Rebased on master

mkoertgen:
@xtreme-shane-lattanzio The jruby build had a hickup with pubkey errors on installing java8:

```console
Reading package lists... Done
W: GPG error: http://ppa.launchpad.net trusty InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY C2518248EEA14886
+ echo 'oracle-java8-installer shared/accepted-oracle-license-v1-1 select true'
+ sudo debconf-set-selections
+ apt -y install oracle-java8-set-default
```

I could add a `--force-yes` on the apt install, however this is not directly related to my PR.
What do yout think?

xtreme-shane-lattanzio:
@mkoertgen Looks like it was a one time thing. Looks good now so merging!

xtreme-shane-lattanzio:
Bumping for webhook trigger.

pivotal-issuemaster:
@mkoertgen Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=499)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=499) to manually synchronize the status of this Pull Request.

PR #511, opened 2018-09-25
Adds Go Module support

osis:


lib/license_finder/package_managers/go_modules.rb
```diff
@@ -0,0 +1,52 @@
...
+      def takes_priority_over
+        Go15VendorExperiment
+      end
+
+      def prepare_command
+        'go mod vendor'
+      end
+    end
+
+    def active?
```
xtreme-vikram-yadav: This method can be tested

PR #527, opened 2019-02-11
allow npm list to return errors

nicholasklem:
npm list returns a variety of errors:
`npm ERR! peer dep missing:...`
But it still manages to list out all the other dependencies. Allthough it would be nice to fix all the errors, I think the license harvesting would benefit from not failing on npm errors. `... | cat ` "filters" out stderr and returns success. If stdout is not json I'm hoping `JSON.parse(stdout)` will crash and burn.


pivotal-pmital:
Can you please update the tests to reflect your changes? The tests are currently failing. Thanks! :)

xtreme-shane-lattanzio:
Moved work to #596

PR #530, opened 2019-02-12
Moves version into its own file

osis:
This fixes a bundler loading dependency loop and expands upon the existing patch version bump job with simpler patch, minor, major CI jobs.


xtreme-vikram-yadav:
Nice refactor!! 

Looks like we don't need the old scripts and tasks anymore :). Specifically, 
https://github.com/pivotal-legacy/LicenseFinder/blob/master/ci/scripts/bump-patch-version.sh
https://github.com/pivotal-legacy/LicenseFinder/blob/master/ci/tasks/bump-patch-version.yml

osis:
Goooood point!

osis:
@xtreme-vikram-yadav removed! 👍

xtreme-shane-lattanzio:
@osis So much better than writing our own updating :D If we are going to do this, we should change how the pipeline works. We used to have the passed code feed into the version bump. That job would only bump patches and then trigger the real release. The manual release job is basically just the regular release job without any patching. I think we should now also remove that manual release job and have the three new jobs take place in that column. We can then have those new jobs trigger the release job. Does that make sense? Let me know if I'm not clear and thanks for cleaning this up!!

osis:
@xtreme-shane-lattanzio Yea, there is a bunch of flow tidy-up that can be done. I just didn't want to feature creep too far away from my PR. I did have a bunch of followups I was planning out.

osis:
@xtreme-shane-lattanzio @xtreme-vikram-yadav Updated. Reflect of new flow can be seen... https://norsk.cf-app.com/teams/main/pipelines/LicenseFinder-test

PR #531, opened 2019-02-13
fix when `bundle install --path` option is used

ericsullivan:
Addresses: https://github.com/pivotal-legacy/LicenseFinder/issues/524

`Bundler.configure` needs to be called before `Bundler::Definition.build` is called (to work when `bundle install --path X` is used). 

For my use case, I installed licence_finder with `gem install licence_finder`. Then, I cloned a repo and ran `bundle install --path vendor/bundle` (I don't want to install the gems globally). Running `licence_finder` was giving an error that the gems weren't installed.


xtreme-shane-lattanzio:
Thanks for the PR :D Your tests seem to be failing CI unfortunately. Please touch it up and we can take a look again

ericsullivan:
hi @xtreme-andy-shen Any ideas what might be going on here? I can get this to pass locally, both with and without bundler. After running `rake install` I can also run it in standalone mode on another repo which has the bundler path set (it fails to do that on master), so I think this code is correct. I assume the other package managers are overlapping ENV (which I can likely fix), but I'm at a loss as to why the tests I added aren't working.

xtreme-shane-lattanzio:
Hi @ericsullivan ! I looked into this a bit and I do see the regular flow working while the tests do not. I think the issue is with the difference of bundle exec vs non bundle exec. I am not exactly sure but it appears that the test run is not using the .bundle file that is in the directory and therefore does not see the rake gem. bundle exec would solve this but then it will say that license_finder is not found because it is not reference in the gemfile. These are just some high level thoughts and I'm not sure if they can spark any progress here...I can try to take a look more in depth when I have a chance but for now, this is all I can see.

osis:
@ericsullivan you can also add `:focus` to [your test](https://github.com/pivotal-legacy/LicenseFinder/pull/531/files#diff-b486c714e38ca4df2c4719e6b3488a00R20) and run `./dlf bundle exec rake features:focus` to test locally.

osis:
Definitely seeing some erratic behaviour as you switch around ruby versions in the docker image. Haven't isolated, but I'm outputting the gem env vars before and after the `Bundler.configure` and when it doesn't work the env vars stay... 
```
GEM_HOME=/usr/local/rvm/gems/ruby-2.6.0
GEM_PATH=/usr/local/rvm/gems/ruby-2.6.0:/usr/local/rvm/gems/ruby-2.6.0@global
```
And when it works, they switch to...
```
GEM_HOME=/scan/myapp/vendor/bundle/ruby/2.6.0
BUNDLER_ORIG_GEM_HOME=/usr/local/rvm/gems/ruby-2.6.0
```

Will dig into Bundler.configure to see what's up.
GEM_PATH=

ericsullivan:
Thank you for looking into this. I dug further into the bundler source code than I ever had and still didn't come up with a solution.

osis:
No worries, this is a weird one.

I dug into Bundler.configure and debugged how it loads, and caches, the config and sets ENV vars. A few interesting notes so consider...

In the Docker image when I run vanilla irb it runs Bundler.configure and no matter what I do with Bundler.with_*_env blocks, I cannot get Bundler.configure to execute again because the configuration is cached. This behaviour works as expected on my local env. 

I can also get the vendored case to pass by running `GEM_PATH='' license_finder`. Which further eludes to Bundler loading a config and caching auto-magically in the LicenseFinder image's env for some reason.  I cannot replicate this automatic execution of Bundler.configure on my local env. Plus the added case that the Bundler.with_*_env methods aren't executing properly.

I feel like there is a minor session issue somewhere. Something along the lines of "Because you're sudo or not instantiating a bash session in a normal way I'm (Bundler) going to act a bit differently or can't do certain things". This is just a hunch though.

osis:
I see. I think there is a cascading effect. I was able to create a predictable outcome, regardless of RVM-ness, via Bundler.reset! and then a Bundler.configure.

But then you run into the `Unable to find a spec satisfying license_finder in the set.` error which is a little bit of a bug(?) since license_finder in the image is scanning itself even though it's not listed as an explicit dependency. The properly bundled env is excluding license_finder, hence the error.

osis:
Figured this out. It's because the Bundler does some voodoo when there is a .bundle in parent directories. Currently when LicenseFinder runs tests, it's executed in the `tmp/` instead of `/tmp/` and you're running into this use case.

@ericsullivan I'll make a PR for the test directory root change, that should fix your issue and help you simplify your code.

osis:
Eventually tracked down the root cause to https://github.com/rvm/rubygems-bundler

```root@a3373ac61abd:/scan/my_app# NOEXEC_DEBUG=1 license_finder
Bin used: /usr/local/rvm/gems/ruby-2.6.1/bin/license_finder
Bin calculated: license_finder
Noexec - starting check
Examining /scan/my_app/Gemfile
Examining /scan/Gemfile```

NOEXEC_DISABLE=1 license_finder fixes the issue when running manually. Might want to dig into the code and see how to undo what it's doing programmatically? 🤔

osis:
Reading [this block](https://github.com/rvm/rubygems-bundler/blob/master/lib/rubygems-bundler/noexec.rb#L27-L67), indicates that we'll only see this issue with a vendored project being downstream of the Gemfile for the binary being executed.

ex. This issue only crops up if we have a vendored project checked out in the license finder directory.

ericsullivan:
Work moved to: https://github.com/pivotal-legacy/LicenseFinder/pull/558

PR #544, opened 2019-02-15
Migrates to the newer Concourse Github Pull Request Resource

osis:
This newer resource will ensure that that pipelines will run on each change, the older one would miss events if they happened in quick succession. 

Also had to rename the resource to avoid https://github.com/telia-oss/github-pr-resource/issues/64.


osis:
This is already live. PR is for consistency. https://norsk.cf-app.com/teams/main/pipelines/LicenseFinder-pr

PR #551, opened 2019-02-20
Syncs Rubocop CI task with bundled version

osis:
There is still a chance that there might be minor experience mismatches
due to the timeliness of a bundle, but this shouldn't happen often at
all.


osis:
This makes auto-PRs like https://github.com/pivotal-legacy/LicenseFinder/pull/548 more relevant as well.

PR #231, opened 2016-06-14
Add support for PHP via Composer

SimantovYousoufov:
Added a `composer` and `composer_package` handlers as well as specs+fixtures for both and a travis config.

I tried to keep with the existing code style but as this is my first dive into the Ruby world so please let me know of any conventions I didn't follow and I'll resolve them ASAP. Would be much appreciated :).

Please note that I added `sudo: required` to `.travis.yml` to get TravisCI to build correctly. There were no whitelisted repositories from which I could use the `apt` option to install PHP. This approach should probably be revisited.

PR #556, opened 2019-02-28
[WIP] Add php support

temp:
Hi, I have re-activated the code from #231 and changed it to fit the latest master.
features and spec are included.
Docker images build runs fine, i have tested it on several of my PHP projects, and it works like a charm.
Since I am not that deep into ruby, any help would be appreciated.
I could provide a minimal php/composer repo on base which the php support for LicenseFinder can be tested.


xtreme-shane-lattanzio:
Hey @temp ! Thanks for restarting this. So it looks like there are 4 test failures with the new composer package manager. Now I don't really know how this thing works but off initial glance it appears that the possible package paths method which is used to detect active package managers should look for the compose json and not the lock. As for the other 3 failures, I am completely lost as that is very much composer stuff and I do not know what it is supposed to be doing. It seems to be stubbing out a `capture` method that I cannot even see. When it runs the `current_packages` method in the composer file, it is failing on the line that is fetching the dependencies from the composer json. 

It almost looks like it is trying to stub a method call that the actual package manager would make and not the class itself which would make no sense. Instead you would stub out the command line run and have it return what you want. Again I don't know how composer works so I am just guessing.

If you could take a look at those and try to shed some light on this it would be helpful. Once we get those tests passing we can look at merging this in.

temp:
Hi @xtreme-shane-lattanzio , thanks for the reply. Will try to figure this out, like I said, the code maily based on the work from the prior php-support PR. And I haven't got a clue regarding Ruby :-D
But I will try to find a colleague with Ruby knowledge and figure this out. 
I will mark this PR as work in progress.

xtreme-shane-lattanzio:
Work moved to #589

PR #558, opened 2019-03-11
Support for Bundler's Vendored Gems

osis:
This is the brain child of @ericsullivan via #531. This PR is just adds some minor adjustments to the overall testing framework and tweaks the implementation a bit.

### DIFF
--- a/ci/scripts/run-tests.sh
+++ b/ci/scripts/run-tests.sh
@@ -19,6 +19,6 @@ pushd "$PROJECT_ROOT"
 
   bundle exec rake install
   bundle exec rake spec
-  bundle exec rake features
 
+  bundle exec rake features
 popd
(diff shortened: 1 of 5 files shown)


ci/scripts/run-tests.sh
```diff
@@ -19,6 +19,6 @@ pushd "$PROJECT_ROOT"
 
   bundle exec rake install
   bundle exec rake spec
-  bundle exec rake features
 
+  rake features
```
xtreme-vikram-yadav: Nitpicking here but it would look better with consistent calls to rake

osis: Not possible if you want to test vendored bundle projects. A `bundle exec` will always the local gem env and using the `with_original_env` method won't work because of RVM's env hacking explained in #531

osis: Actually, let me try one thing...

xtreme-vikram-yadav: Got it.

xtreme-vikram-yadav:
Oh man, i was a second away from merging it ;)

osis:
Hahaha I found a way around it... Should be good to go! 🎉

ericsullivan:
congratulations @osis That was a tough PR. I really appreciate you working through it.

PR #559, opened 2019-03-13
Widens options for Dockerfile build triggering

osis:


osis:
There are two conclusions to follow up on...

1. Build every time there is a commit to master
2. Leave it tied to dependency change inducing files (this PR) and tighten up versioning in the Gemfile.

PR #567, opened 2019-03-29
Revert "Revert "Fixes bundler env configuration for vendored gems""

xtreme-vikram-yadav:
This reverts commit 98b2bb8.

Reverts previous reverts:
Revert "Fixes bundler env configuration for vendored gems" (commit: 6b9d1ea)
Revert "fix when bundle install --path option is used" (commit: 8c4e677)

### DIFF
--- a/ci/scripts/run-tests.sh
+++ b/ci/scripts/run-tests.sh
@@ -19,6 +19,6 @@ pushd "$PROJECT_ROOT"
 
   bundle exec rake install
   bundle exec rake spec
-  bundle exec rake features
 
+  bundle exec rake features
 popd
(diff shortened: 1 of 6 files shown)


xtreme-shane-lattanzio:
FYI this caused the systems tests for OSLO to fail when running license finder on the web directory in the **norsk-sample-release** repo in the **pivotal-cf-experimental** github org. To repro you will need to be on the **new-system-tests** branch in that repo and simply run `license_finder --prepare` in the **web** directory

osis:
This fixes https://github.com/pivotal-legacy/LicenseFinder/issues/579

PR #577, opened 2019-05-02
Switches from Oracle JDK to Open JDK

osis:


osis:
Pushing this through as an incremental fix to the overall CI issue that solves https://github.com/pivotal-legacy/LicenseFinder/pull/572#issuecomment-486753049 & https://norsk.cf-app.com/teams/main/pipelines/LicenseFinder/jobs/build-docker-image/builds/2

PR #580, opened 2019-05-04
Updates test version of http-server to something valid for YARN and NPM

### LINKED ISSUE #578: Expired NPM Module is Holding Up CI
Looks like the version of http-server used in the feature tests needs to be bumped.

osis:
Fixes #578


osis:
I think this version of http-server is back up, or maybe it was a connection issue with the CI cluster? Either way, the pipeline has gone green without this change. Might not hurt to bump this version regardless.

PR #589, opened 2019-05-24
Add composer PHP support

knightzac19:
Using the code from #556 I was able to get master merged in and tests working. Tested on a Yii2 PHP codebase with over 900 dependencies between PHP and Yarn without any issues.


kaystrobach:
Do i get it right, that composer is then needed as dependency? Wasn’t the license information available in the lock file?

knightzac19:
I just implemented it like most of the other implementations, there is license info in the lock file, but it's just nested under packages and packages-dev.

kaystrobach:
Ok, so no need for composer - perfect!

knightzac19:
No, you'll still need to have composer, the way this works is it's calling composer's license call to get all the licenses for the packages you have installed in a simple json object array. If you use the docker image, it will have the necessary packages to run composer checks.

spec/lib/license_finder/package_managers/composer_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+        'require' => {
+          'vlucas/phpdotenv' => '3.3.*'
+        },
+        'require-dev' => {
+          'symfony/debug' => '4.2.*'
+        }
+      }.to_json
+    end
+
+    describe '.prepare' do
```
xtreme-shane-lattanzio: Looks like the current_packages test was removed? Why was this? It probably should be added in addition to the prepare test.

spec/lib/license_finder/package_managers/composer_package_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    its(:name) { should == 'symfony/debug' }
+    its(:version) { should == 'v3.0.7' }
+    its(:summary) { should eq '' }
+    its(:description) { should == '' }
+    its(:homepage) { should == '' }
+    its(:package_manager) { should eq 'Composer' }
+
+    describe '#license_names_from_spec' do
+      it 'finds the license for both license structures' do
```
xtreme-shane-lattanzio: I don't know much about composer but in #556 there were additional test cases here to handle different formats. Why was this removed?

Dockerfile
```diff
@@ -9,6 +9,7 @@ ENV GRADLE_VERSION 4.10.3
 ENV RUBY_VERSION 2.6.1
 ENV MIX_VERSION 1.0
 ENV JDK_VERISON 8u211
+ENV COMPOSER_ALLOW_SUPERUSER 1
```
xtreme-shane-lattanzio: Curious as to why this is needed?

knightzac19: There's a warning that gets outputted on the docker image that causes invalid json. Adding that silences the composer warning.

spec/lib/license_finder/package_managers/composer_package_spec.rb
knightzac19: Those were all taken from NPM's test suite and not relevant to composer.

spec/lib/license_finder/package_managers/composer_spec.rb
knightzac19: Looks like I overlooked that because tests were failing so bad, I will see how hard it is to put current_packages check back in.

knightzac19: I've added back in the tests and fixed a couple issues that popped up from it, should be good to go once the pipeline clears.

ci/pipelines/pull-request.yml.erb
```diff
@@ -50,6 +50,11 @@ jobs:
     - put: dockerhub-pr
       params:
         build: github-pull-request
+    - put: github-pull-request
```
xtreme-shane-lattanzio: I think this was a bad rebase. This should be reverted.

knightzac19: Not sure what happened there, I reverted that last commit.

xtreme-shane-lattanzio:
Bumping for webhook trigger.

pivotal-issuemaster:
@knightzac19 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=589)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=589) to manually synchronize the status of this Pull Request.

PR #576, opened 2019-05-01
continue on unmet peer dependency

antjkennedy:
* failing where we have an unmet peer dependency, but should still be possible to produce report

PR #596, opened 2019-06-03
[Changed] NPM will now output proper errors

xtreme-shane-lattanzio:
This fixes the errors that were initially worked on in #527 by @nicholasklem


xtreme-shane-lattanzio:
Closing this in favour of #576 . The raise would never execute. The other way will continue if there are certain errors and raise otherwise

PR #598, opened 2019-06-03
Updated dockerfile to stop GPG failures on rvm step

knightzac19:
I noticed the build-docker-image step for the CI (and on my local) liked to fail a lot on the gpg step for RVM, took a peak at RVM and noticed that they have a PPA method for ubuntu to use instead of the script route. I've updated the Dockerfile to use that instead and have not seen any more failures.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,9 @@ RUN apt-get update && apt-get install -y \
   git-core \
   sudo \
   unzip \
-  wget
+  wget \
+  gnupg2 \ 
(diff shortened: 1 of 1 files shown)


knightzac19:
@xtreme-shane-lattanzio Would you be up to merging this in once passed? I'm already seeing a bunch of CI jobs fail because of the RVM gpg key failing to pull in.

xtreme-shane-lattanzio:
@knightzac19 Thank you again!

PR #599, opened 2019-06-03
continue on unmet peer dependency

xtreme-shane-lattanzio:
Work initially on #576 by antjkennedy


pivotal-so:
Merging; thanks for the work @antjkennedy!

PR #604, opened 2019-07-05
Make yarn prepare_command use --ignore-engines flag

jacknewberry:
### Background
`package.json` can contain an "engines" key that can be used to specify the runtime versions that the project should run on.
e.g:
```
"engines": {
    "node": ">= 0.11.16",
    "npm" : "~1.0.20"
}
```

When `yarn install` is run, yarn checks that the local runtimes match
the specified versions.

### Problem
`yarn install` will fail if a specified engine version is not found. This causes LicenseFinder to fail. This behaviour is undesirable because the failure is not related to finding the project's licenses.

### Solution
The failure can be avoided by running yarn with the `--ignore-engines` flag.
This flag has been available since yarn 0.13.0, Sept 2016:
https://github.com/yarnpkg/yarn/commit/a0b8195c012eda24ce4ea8b9da60dce29794012a


xtreme-vikram-yadav:
@jacknewberry I'm trying to understand the ramifications after this change goes through. There are cases when we might want to use this option but most cases, there may be other alternatives than using this flag like either using the proper engine or reinstall dependencies. Also, when license_finder is running, I'd assume the environment already have package managers (with right versions for engines) required to run prepare commands. But in case this is absolutely the only option, I'd expect this to be an option provided by the license_finder CLI and not introduced for all projects.

jacknewberry:
I think that the purpose of LicenseFinder is to find licenses. It uses tools like Yarn and Maven to help find licences.
Licence finder is not a tool for detecting whether the project has the correct engine.

Yarn performs the engine check because it is common that immediately after running `yarn install`, the user wants to build and run their app. This is not the case when Yarn is being used by LicenseFinder. The engine check in this context is an additional test that is not required for finding licenses.

Checking that the local machine has the correct engine installed would be similar to if Yarn checked there is enough memory available for the app to run. This might be a useful feature for Yarn, but it would be undesirable for LicenseFinder.

PR #627, opened 2019-10-02
Remove gradle subprojects from "project_roots" list

xtreme-jason-smith:
Gradle submodules are no longer reported as project_roots if there are no other package manager files located in the submodule root folder.


lib/license_finder/project_finder.rb
```diff
@@ -6,6 +6,7 @@ def initialize(main_project_path, strict_matching = false)
       @package_managers = LicenseFinder::Scanner::PACKAGE_MANAGERS
       @strict_matching = strict_matching
       @main_project_path = main_project_path
+      @gradle_subprojects_paths = []
```
xtreme-jason-smith: This needs to be removed, as it's no longer used.

Petahhh:
Please don't forget to squash the commits (or rename them) because  I see a wip commit in there.

lib/license_finder/package_managers/gradle.rb
```diff
@@ -42,8 +42,26 @@ def package_management_command
       File.exist?(File.join(project_path, wrapper)) ? wrapper : gradle
     end
 
+    def active?
+      result = super
+
+      # If it is a valid gradle module, is it a root module?
+      result = root_module? if result
```
xtreme-vikram-yadav: My only concern around this check is that if the user is inside a gradle subproject, running license_finder will say it does detect gradle which can be confusing for users. If we are going with this change, i would recommend a warning message describing why license_finder did not recognize gradle package manager inside a gradle subproject directory.

spec/fixtures/gradle-with-subprojects/gradlew.bat
```diff
@@ -0,0 +1,84 @@
+@if "%DEBUG%" == "" @echo off
```
xtreme-vikram-yadav: I'm not sure if we need .bat file included for tests

lib/license_finder/package_managers/gradle.rb
xtreme-vikram-yadav: Also, i would recommend confirming with Kim as this is a little bit out of scope of the story.

spec/lib/license_finder/package_managers/gradle_spec.rb
```diff
@@ -147,46 +148,83 @@ module LicenseFinder
...
       include FakeFS::SpecHelpers
 
-      it 'return true if build.gradle exists' do
-        FakeFS do
-          FileUtils.mkdir_p '/fake/path'
-          FileUtils.touch '/fake/path/build.gradle'
+      context 'when dealing with root gradle project' do
+        context "when there's a build.gradle" do
+          it 'returns true' do
+            FakeFS do
```
xtreme-vikram-yadav: having setup blocks might save some nested code here: 
```
before do
    FakeFS.activate!
  end

  after do
    FakeFS.deactivate!
  end
```

spec/lib/license_finder/package_managers/gradle_spec.rb
```diff
@@ -147,46 +148,83 @@ module LicenseFinder
...
+        context "when there's no build.gradle or build.gradle.kts" do
+          it 'returns false' do
+            expect(SharedHelpers::Cmd).not_to receive(:run).with("gradle properties | grep 'parent: '")
+            expect(subject.active?).to be false
+          end
+        end
+
+        context "when there's build.gradle.kts" do
+          it 'return true' do
+            FakeFS do
```
xtreme-vikram-yadav: Same as above

spec/lib/license_finder/package_managers/gradle_spec.rb
```diff
@@ -147,46 +148,83 @@ module LicenseFinder
...
+          end
+        end
+
+        context "when there's a settings.gradle" do
+          it 'uses the build.gradle referenced inside' do
+            SETTINGS_DOT_GRADLE = <<-GRADLE
+  rootProject.buildFileName = 'build-alt.gradle'
+            GRADLE
+
+            FakeFS do
```
xtreme-vikram-yadav: Same as above

spec/lib/license_finder/package_managers/gradle_spec.rb
```diff
@@ -147,46 +148,83 @@ module LicenseFinder
...
+              expect(SharedHelpers::Cmd).to receive(:run).with("gradle properties | grep 'parent: '").and_call_original
+
+              expect(subject.active?).to be true
+            end
+          end
+        end
+
+        context "when there's a settings.gradle" do
+          it 'uses the build.gradle referenced inside' do
+            SETTINGS_DOT_GRADLE = <<-GRADLE
```
xtreme-vikram-yadav: heredoc looks weird for a single line string

spec/lib/license_finder/package_managers/gradle_spec.rb
```diff
@@ -147,46 +148,83 @@ module LicenseFinder
...
+            end
+          end
         end
       end
 
-      context "when there's build.gradle.kts" do
-        it 'return true' do
+      context 'when dealing with a gradle subproject' do
+        it 'returns false' do
           FakeFS do
```
xtreme-vikram-yadav: Same as above

spec/lib/license_finder/package_managers/gradle_spec.rb
xtreme-jason-smith: I believe this is automatically handled by `include FakeFS::SpecHelpers` which is included in the `#activate?` describe block (line 149)

xtreme-jason-smith:
This PR has been replaced with the following PR #629

PR #629, opened 2019-10-09
Exclude Gradle subprojects from project roots

Pivotal-Jeff-Jun:
Excluding gradle subprojects when running `license_finder project_roots`


README.md
```diff
@@ -424,6 +424,8 @@ downloadLicenses {
 }
 ```
 
+**A note on Gradle Subprojects**: `license_finder` does not report Gradle subprojects as project roots when generating reports. However, if there is a non-Gradle package definition file in a subproject, then it will be included in the report.
```
xtreme-jason-smith: This needs to be updated, as this change only affects 'project_roots' and not 'reports'

lib/license_finder/cli/main.rb
```diff
@@ -91,7 +91,13 @@ def self.shared_options
       shared_options
       def project_roots
         config.strict_matching = true
-        aggregate_paths
+        paths = aggregate_paths
+        filtered_project_roots = Scanner.remove_subprojects(paths)
+        if filtered_project_roots.empty?
+          say(paths)
```
xtreme-jason-smith: So, we are saying that if every path is a subproject, 'project_roots' will still print out all paths, including subprojects?  Isn't that a bit confusing especially if there is at least one project that is a root project, all subprojects will be filtered out in the output.

spec/lib/license_finder/package_managers/gradle_spec.rb
```diff
@@ -147,46 +148,138 @@ module LicenseFinder
...
+              File.open('/fake/path/settings.gradle', 'w') do |file|
+                file.write SETTINGS_DOT_GRADLE
+              end
+              FileUtils.touch '/fake/path/build-alt.gradle'
+
+              expect(SharedHelpers::Cmd).to receive(:run).with("gradle properties | grep 'parent: '").and_call_original
+
+              expect(subject.project_root?).to be true
+            end
+          end
```
xtreme-jason-smith: It looks like this set of tests is very similar (or duplicate) of the 'active?' tests. Is there a way to used shared tests for these two cases?

PR #633, opened 2019-10-24
Create output dir if not existent

mobilutz:
We where running into this error on our CI:
`Errno::ENOENT: No such file or directory @ rb_sysopen - ./audit/licenses.html`

As we are using `--save=./audit/licenses.html` and the directory is not present, this failed.

With this change, the needed directories will be created even when there are multiple ones, i.e. `--save=./audit/in/sub/folder/licenses.html` will create all needed directories.


mobilutz:
@xtreme-shane-lattanzio Thanks for the fixup.

PR #650, opened 2019-12-12
Replace whitelist/blacklist terminology

grantbdev:
The goal of this pull request is to replace whitelist/blacklist terminology in the user-facing functionality and in the codebase. It is a good opportunity to clarify the commands since the old terms are not inherently descriptive. https://github.com/rubocop-hq/rubocop/pull/6464 has links to the original motivation and many changes to other gems that have been replacing those words.

This PR introduces breaking changes. The existing subcommands are completely replaced and the output and the expected input for dependency decision files will be different. Users with an existing dependency decisions file will need to find and replace all instances of `blacklist` with `restrict` and `whitelist` with `permit`.


kdykeman:
Hi @grantbdev, I found [a similar issue related](https://github.com/rails/rails/issues/33677) related to rails documentation. In that case it looks like they went with the terms `permitted` and `restricted`. Is it ok if we go with those terms?

Also, I know it would be a breaking change for some, but what do you think about doing a find/replace on the existing terms rather than supporting two alternatives (one of which might be deemed offensive)?

grantbdev:
> Hi @grantbdev, I found [a similar issue](https://github.com/rails/rails/issues/33677) related to rails documentation. In that case it looks like they went with the terms `permitted` and `restricted`. Is it ok if we go with those terms?

@kdykeman I'm a fan of using `permit` and `permitted` in many cases. However, do you think in this usage it may be confusing: `license_finder permitted add MIT`? Using permitted by itself as a noun sounds weird to me. `permitted_licenses` and `restricted_licenses` is very clear, but longer. Just some things to consider so we can get it right.

> Also, I know it would be a breaking change for some, but what do you think about doing a find/replace on the existing terms rather than supporting two alternatives (one of which might be deemed offensive)?

That sounds good to me. Do you want me to update this PR to change everything at once as a breaking change for a v6?

kdykeman:
> @kdykeman I'm a fan of using `permit` and `permitted` in many cases. However, do you think in this usage it may be confusing: `license_finder permitted add MIT`? Using permitted by itself as a noun sounds weird to me. `permitted_licenses` and `restricted_licenses` is very clear, but longer. Just some things to consider so we can get it right.

@grantbdev I agree it's better to be more verbose to help with clarity. `permitted_licenses` and `restricted_licenses` sounds like the better way to go.

> That sounds good to me. Do you want me to update this PR to change everything at once as a breaking change for a v6?

Yes please... that would be super helpful!

grantbdev:
@kdykeman Thank you for your guidance. PR updated!

pivotal-issuemaster:
@grantbdev Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=650)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=650) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@grantbdev Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=650)!

PR #651, opened 2019-12-13
[FEATURE] Ensure composer always installs the packages

kaystrobach:
This patch ignores some platform requirements to ensure that the packages can be installed.
To achieve that all plugins and all the php version and extension requirements are ignored.

Otherwise a simple missing `php-gd` extension might break the license finder


xtreme-shane-lattanzio:
Hey @kaystrobach ! Thanks for the PR. Can you fix up the tests and then we can have this merged? Thanks!!!

Issue #466, opened 2018-03-14
Add support for pipenv


andrlik:
Given that `pipenv` is now a [Python.org recommended tool](https://packaging.python.org/tutorials/managing-dependencies/#managing-dependencies) for managing project dependencies, it would be good to add support for parsing project Pipfiles.

Issue #230, opened 2016-06-13
Support for private pypi


wayne-luminal:
Looking at pip.rb, pypi is contacted directly for package license information. If you host your libraries on a private pypi (or anything other than pypi.python.org), your private projects will never be found. Is there an option to provide a pypi-like url?

Alternatively, what do you think about using `pip show` for grabbing license information instead of pypi? The good thing about pypi is it's json and predictable but the output of `pip show` is not :/

flavorjones:
I'm open to making a change like this, particularly if we can detect resolved package dependencies without making external network connections.

There's a broader topic here that spans more than just `pypi` -- currently `bundler` and some other package manager classes also go to the network. I'd love if we had the option of examining files on disk whenever possible (see how the `GoDep` class directly parses the `Godeps.json` file for an example of what I mean).

Would you be willing to look into how hard it would be to accomplish this in the `pip` world?

wayne-luminal:
@flavorjones I forked the project and made an attempt to solve this. You can see the diff at: https://github.com/pivotal/LicenseFinder/compare/master...wayne-luminal:pip-instead-of-pypi

I followed the pattern of shelling out to a python script because I don't know much about Ruby but more importantly, I can use the pip classes (I don't know if this is possible in Ruby). The script uses regex to find all the "key: value" pairs and create a json document out of it. For example, `License: Apache License 2.0` becomes `"license": "Apache License 2.0"` in the document. We already know using the pip classes are a little fragile (#224), so I'm not crazy about this approach, but it gets the tool away from reaching out to the network. It might be possible to use `pip show <package-name>` instead. 

One note, if a value (in the key/value output of package metadata) contains a multiline string, only the first line will be retrieved. I've seen some projects where the entire contents of a license file is dumped into place rather than the title of the license itself. For an example, see v0.9.5 of github3.py. This will be fixed in the 1.0 release AFAIK but it is something to consider.

There are still tests to add which I hope to do soon. I wanted to get your initial feedback on this so far.

flavorjones:
I'll try to take a look today. Thanks for the effort!

bikebilly:
Hi @flavorjones, are you still interested to implement the proposed fix by @wayne-luminal?
This seems a very interesting feature to support air-gapped environments with limited internet access.

Thanks!

PR #659, opened 2020-01-09
Add support for Pipfile.lock

mokhan:
This change adds minimal support for `Pipfile.lock`. At the moment [Pipfile.lock](https://github.com/pypa/pipfile#pipfilelock) is generated by `pipenv` with support expected to be added natively to `pip` eventually.

For this change, I did not install the `pipenv` package manager because parsing the `Pipfile.lock` is pretty straightforward and faster. The `pip` package manager class was already making output HTTP requests to `pypi.org` to fetch license information. I extracted this HTTP request generation into a separate utility class and added tests for different scenarios such as connection timeouts and an infinite redirect loop.

I was tempted to also add open and read timeouts but I chose to omit that for now and stick with the default 60 second timeouts that `net/http` ships with.


lib/license_finder/package_managers/pipenv.rb
```diff
@@ -0,0 +1,33 @@
...
+
+module LicenseFinder
+  class Pipenv < PackageManager
+    def initialize(options = {})
+      super
+      @lockfile = Pathname('Pipfile.lock')
+    end
+
+    def current_packages
+      dependencies['default'].map do |name, data|
```
mokhan: By default this is only parsing the `default` dependencies. I'm unsure if I should also parse out the `develop` dependencies as well. WDYT?

lib/license_finder/package_utils/pypi.rb
```diff
@@ -0,0 +1,40 @@
...
+      Net::ReadTimeout,
+      OpenSSL::OpenSSLError,
+      OpenSSL::SSL::SSLError,
+      SocketError,
+      Timeout::Error
+    ].freeze
+
+    class << self
+      def definition(name, version)
+        response = request("https://pypi.org/pypi/#{name}/#{version}/json")
```
mokhan: I'm tempted to extract `pypi.org` as a source parameter so that this option can be customized to point to a `pypi.org` mirror.
Something like:

```ruby
def definition(name, version, source: 'pypi.org')
```
as a possible starting point. Then eventually connecting a CLI option to utilize this.

lib/license_finder/package_utils/pypi.rb
```diff
@@ -0,0 +1,40 @@
...
+        {}
+      end
+
+      def request(location, limit = 10)
+        uri = URI(location)
+        http = Net::HTTP.new(uri.host, uri.port)
+        http.use_ssl = true
+        response = http.get(uri.request_uri).response
+        response.is_a?(Net::HTTPRedirection) && limit.positive? ? request(response['location'], limit - 1) : response
+      end
```
mokhan: I was tempted to add [net-hippie](https://mokhan.ca/%F0%9F%92%8E/2018/11/10/net-hippie-0-2-0.html) as a project dependency which includes retry logic with exponential backoff + jitter. I opted to leave the `net/http` code as is and added tests for this code.

spec/fixtures/pipenv-with-lockfile/Pipfile
```diff
@@ -0,0 +1,12 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
```
mokhan: I didn't add an example `dev` package but looking at this now I'm think I should. WDYT?

xtreme-shane-lattanzio:
Relevant issue: https://github.com/pivotal/LicenseFinder/issues/466

lib/license_finder/package_managers/pipenv.rb
```diff
@@ -0,0 +1,63 @@
...
+        end
+      end
+    end
+
+    def canonicalize(version)
+      version.sub(/^==/, '')
+    end
+
+    def build_package_for(name, version)
+      PipPackage.new(name, version, PyPI.definition(name, version))
```
mokhan: I don't like having to make an individual API call for each package. Does anyone know if the Pypi has the equivalent of a batch API endpoint to fetch all the data in one call?

lib/license_finder/package_utils/pypi.rb
xtreme-shane-lattanzio: Is this something that people actually do? I don't mind either way but for a first pass at least this is fine.

mokhan:
> Thanks so much for the PR! Looks good. Only request from me is if you could add a feature test to cover this? The tests you have already written are good but doesn't hurt to cover more! You should be able to just add more to the testing dsl file and print what the expected output is similar to the other pip feature spec.

Ah, yes of course!

lib/license_finder/package_utils/pypi.rb
mokhan: Yep. In some organizations packages can only be pulled from an internal mirror where package analysis can be performed on the mirror. This is also true in some air gap scenarios.

Python even has a guide on setting up a mirror. https://packaging.python.org/guides/index-mirrors-and-caches/#id5

I'm going to skip this change because it's a little more nuanced than specifying a single source. The `Pipfile.lock` meta section can actually specify multiple sources to try to pull a package from.

As a starting point, I think defaulting to `pypi.org` is probably a good starting point. Also, this PR is getting rather large.

pivotal-issuemaster:
@mokhan Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=659)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=659) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@mokhan Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=659)!

xtreme-shane-lattanzio:
Fox context, there is an existing issue to support private pypi which this PR lays groundwork for: https://github.com/pivotal/LicenseFinder/issues/230

PR #660, opened 2020-01-10
Add ruby 2.7.0 and update to latest patch levels

mokhan:
This change updates the supported rubies. It adds `2.7.0` and updates the other already supported rubies to the latest ruby patch level.


xtreme-shane-lattanzio:
I know that this PR is still in draft state but in order to support this we will probably also want to bump the ruby version in the Dockerfile to be 2.7.0.

xtreme-shane-lattanzio:
Had to revert the actual Dockerfile change because I forgot about our internal tools syncing with this. We will upgrade once that is done but will leave the tests running on latest!

PR #697, opened 2020-03-22
[FIXED] Replace toml parser with tomlrb

mvz:
The toml gem depends on an old version of parslet, which can cause conflicts if license_finder is included in a Gemfile. On the other hand, tomlrb has no extra dependencies and seems more maintained.


xtreme-shane-lattanzio:
Hey @mvz thanks for the PR. Can you please take a look at the error in the pipeline. It looks like something is different with the toml parsing due to this change.

PR #703, opened 2020-03-31
[Changed] Run glide install in folder containing glide.lock

xtreme-jason-smith:
When glide.lock file is located inside src sub directory, run glide
install in that directory instead of project root directory.

[#171957675]

### DIFF
--- a/features/features/package_managers/glide_spec.rb
+++ b/features/features/package_managers/glide_spec.rb
@@ -7,22 +7,35 @@ describe 'Glide Dependencies' do
 
   context 'when project is in src directory' do
     specify 'are shown in reports for a project' do
-      project = LicenseFinder::TestingDSL::GlideProject.create
-      ENV['GOPATH'] = "#{project.project_dir}/gopath_glide"
+      LicenseFinder::TestingDSL::GlideProject.create
 
-      go_developer.run_license_finder('gopath_glide')
+      go_developer.run_license_finder('src/gopath_glide/src')
       expect(go_developer).to be_seeing_line 'github.com/Masterminds/semver, 517734cc7d6470c0d07130e40fd40bdeb9bcd3fd, MIT'
       expect(go_developer).to be_seeing_line 'github.com/Masterminds/cookoo, 78aa11ce75e257c51be7ea945edb84cf19c4a6de, MIT'
     end
   end
 
-  context 'when projecy is not in src directory' do
+  context 'when project in root directory' do
     specify 'are shown in reports for a project' do
       LicenseFinder::TestingDSL::GlideProjectWithoutSrc.create
 
-      go_developer.run_license_finder('gopath_glide_without_src')
+      go_developer.run_license_finder('src/gopath_glide_without_src')
       expect(go_developer).to be_seeing_line 'github.com/Masterminds/semver, 517734cc7d6470c0d07130e40fd40bdeb9bcd3fd, MIT'
       expect(go_developer).to be_seeing_line 'github.com/Masterminds/cookoo, 78aa11ce75e257c51be7ea945edb84cf19c4a6de, MIT'
+    end
+  end
+
+  context 'when project in both root and src directory' do
+    specify 'are shown in reports for a project' do
+      LicenseFinder::TestingDSL::GlideProjectWithRootAndSrc.create
+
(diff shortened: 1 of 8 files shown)


pivotal-issuemaster:
@xtreme-jason-smith Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=703)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=703) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@xtreme-jason-smith Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=703)!

lib/license_finder/package_managers/glide.rb
```diff
@@ -35,7 +35,10 @@ def package_management_command
     end
 
     def prepare_command
-      'glide install'
+      command = 'glide install'
+      command = "cd src; #{command}; cd .." if File.exist?(project_path.join('src', 'glide.lock'))
```
pivotal-pmital: @xtreme-jason-smith Is it possible that both `project_path` and `project_path/src` may have the `glide.lock` file? If it is possible, how do we want to handle that?

If we know for sure that it never happens, feel free to Resolve this comment. The rest looks good to me

xtreme-shane-lattanzio: Yea i'm not sure if this is a safe assumption to make. The proper way to do this would be to make sure the --project_path argument is being used properly. It should install the glide command using that path. We really shouldn't be changing directories in the prepare commands.

Pivotal-Jeff-Jun: @xtreme-shane-lattanzio - Yes, that would make sense. However, the glide package manager checks both the root folder and the `src` sub folder to determine if it's an active glide project. So, to do your suggestion, we would need to remove the `src` folder check, which would revert the change where certain projects put their glide lock files in the `src` folder instead of the root project.

I guess we could ask tracker to fix their glide project to host those files in the root folder instead?  However, I wonder if this is common practice for glide, meaning the `glide.lock` file are not always in the base directory.

Pivotal-Jeff-Jun: @pivotal-pmital We could run it in both folders if we find the lock file in both. Though like you I'm not sure how common that would be.

Also, as mentioned in reply to Shane, is the team that has this issue created their glide project correct? Maybe the glide.lock file should not be in the `src` folder, but in the base folder.

xtreme-vikram-yadav:
`https://github.com/pivotal/LicenseFinder/blob/ba36bb7505e6dfb285dc10a64dd8e8bbc2088d3f/features/features/package_managers/glide_spec.rb#L8` needs to be updated to reflect this change as well.

features/support/testing_dsl.rb
```diff
@@ -98,6 +98,10 @@ def project_dir
         Paths.project(name)
       end
 
+      def cloneGoProject(fixture_name)
```
xtreme-shane-lattanzio, line 101: I don't think you need a separate function for this. It should be a part of the fixture itself.

features/features/package_managers/glide_spec.rb
```diff
@@ -7,22 +7,35 @@
...
     end
   end
 
-  context 'when projecy is not in src directory' do
+  context 'when project in root directory' do
     specify 'are shown in reports for a project' do
       LicenseFinder::TestingDSL::GlideProjectWithoutSrc.create
 
-      go_developer.run_license_finder('gopath_glide_without_src')
+      go_developer.run_license_finder('src/gopath_glide_without_src')
```
xtreme-shane-lattanzio, line 22: Same question about the parent src directory

xtreme-jason-smith: All those changes is to reflect a proper go project setup.  GOPATH must have a `src` subdirectory or else glide will not work.  And all project usually are located inside the GOPATH/src directory.  So, these changes reflect a better GO environment setup for testing, and is closer to what NORSK does as well.

Issue #717, opened 2020-05-03
Docker licensefinder/license_finder seems to use old PHP version, upgrade to 7.3 or 7.4.


Yivan:
Hello,

First, thanks a lot for this very nice tools shared with the community.

Running `docker run -t --rm licensefinder/license_finder:latest ...` was working great until Ocramius php package abandonned PHP 7.2 support.
I get this bug when trying to generate licenses for a PHP project:

```
Digest: sha256:db4b2c335af7cff297f14385ddc4dcdea87033a726c2943b941cec15ed721f94
Status: Downloaded newer image for licensefinder/license_finder:latest
Generate License Finder report...
Traceback (most recent call last):
        22: from /usr/share/rvm/gems/ruby-2.6.5/bin/license_finder:23:in `<main>'
        21: from /usr/share/rvm/gems/ruby-2.6.5/bin/license_finder:23:in `load'
        20: from /usr/share/rvm/gems/ruby-2.6.5/gems/license_finder-6.1.2/bin/license_finder:6:in `<top (required)>'
        19: from /usr/share/rvm/gems/ruby-2.6.5/gems/thor-1.0.1/lib/thor/base.rb:485:in `start'
        18: from /usr/share/rvm/gems/ruby-2.6.5/gems/thor-1.0.1/lib/thor.rb:392:in `dispatch'
        17: from /usr/share/rvm/gems/ruby-2.6.5/gems/thor-1.0.1/lib/thor/invocation.rb:127:in `invoke_command'
        16: from /usr/share/rvm/gems/ruby-2.6.5/gems/thor-1.0.1/lib/thor/command.rb:27:in `run'
        15: from /usr/share/rvm/gems/ruby-2.6.5/gems/license_finder-6.1.2/lib/license_finder/cli/main.rb:147:in `report'
        14: from /usr/share/rvm/gems/ruby-2.6.5/gems/license_finder-6.1.2/lib/license_finder/license_aggregator.rb:11:in `dependencies'
        13: from /usr/share/rvm/gems/ruby-2.6.5/gems/license_finder-6.1.2/lib/license_finder/license_aggregator.rb:49:in `aggregate_packages'
        12: from /usr/share/rvm/gems/ruby-2.6.5/gems/license_finder-6.1.2/lib/license_finder/license_aggregator.rb:49:in `flat_map'
[...]

PR #718, opened 2020-05-04
[Changed] Bump PHP version to 7.3

xtreme-shane-lattanzio:


xtreme-shane-lattanzio:
Related: https://github.com/pivotal/LicenseFinder/issues/717

PR #721, opened 2020-05-08
Made LF license matching more dynamic

Pivotal-Jeff-Jun:
* made license detection from templates more flexible 

* updated Simplifed BSD license

[#171961625]

### DIFF
--- a/lib/license_finder/license/text.rb
+++ b/lib/license_finder/license/text.rb
@@ -11,21 +11,30 @@ module LicenseFinder
       ALPHABET_ORDERED_LIST = /\\\([a-z]\\\)\\\s/.freeze
       ALPHABET_ORDERED_LIST_OPTIONAL = '(\([a-z]\)\s)?'
       LIST_BULLETS = /(\d{1,2}\\\.|\\\*)\\\s/.freeze
-      LIST_BULLETS_OPTIONAL = '(\d{1,2}.|\*)\s*'
+      LIST_BULLETS_OPTIONAL = '(\d{1,2}.|\*)?\s*'
+      NEWLINE_CHARACTER = /\n+/.freeze
+      QUOTE_COMMENT_CHARACTER = /^\s*\>+/.freeze
+      ESCAPED_QUOTES = /\\\"/.freeze
 
       def self.normalize_punctuation(text)
         text.dup.force_encoding('UTF-8')
-                 .gsub(SPECIAL_DOUBLE_QUOTES, '"')
-                 .gsub(SPECIAL_SINGLE_QUOTES, "'")
-                 .gsub(SPACES, ' ')
-                 .gsub(QUOTES, '"')
-                 .strip
+            .gsub(SPECIAL_DOUBLE_QUOTES, '"')
+            .gsub(SPECIAL_SINGLE_QUOTES, "'")
+            .gsub(QUOTE_COMMENT_CHARACTER, '')
+            .gsub(SPACES, ' ')
+            .gsub(NEWLINE_CHARACTER, ' ')
+            .gsub(ESCAPED_QUOTES, '"')
+            .gsub(QUOTES, '"')
+            .strip
+      rescue ArgumentError => _e
+        text
       end
 
       def self.compile_to_regex(text)
         Regexp.new(Regexp.escape(normalize_punctuation(text))
                        .gsub(PLACEHOLDERS, '(.*)')
                        .gsub(',', '(,)?')
+                       .gsub('HOLDER', '(HOLDER|OWNER)')
                        .gsub(ALPHABET_ORDERED_LIST, ALPHABET_ORDERED_LIST_OPTIONAL)
                        .gsub(LIST_BULLETS, LIST_BULLETS_OPTIONAL))
       end
--- a/lib/license_finder/license/text.rb
+++ b/lib/license_finder/license/text.rb
@@ -6,15 +6,28 @@ module LicenseFinder
       SPACES = /\s+/.freeze
       QUOTES = /['`"]{1,2}/.freeze
       PLACEHOLDERS = /<[^<>]+>/.freeze
+      SPECIAL_SINGLE_QUOTES = /[‘’]/.freeze
+      SPECIAL_DOUBLE_QUOTES = /[“”„«»]/.freeze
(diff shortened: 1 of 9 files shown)


lib/license_finder/license/text.rb
```diff
@@ -6,15 +6,28 @@ module Text
       SPACES = /\s+/.freeze
       QUOTES = /['`"]{1,2}/.freeze
       PLACEHOLDERS = /<[^<>]+>/.freeze
+      SPECIAL_SINGLE_QUOTES = /[‘’]/.freeze
+      SPECIAL_DOUBLE_QUOTES = /[“”„«»]/.freeze
+      ALPHABET_ORDERED_LIST = /\\\([a-z]\\\)\\\s/.freeze
+      ALPHABET_ORDERED_LIST_OPTIONAL = '(\([a-z]\)\s)?'
+      LIST_BULLETS = /(\d{1,2}\\\.|\\\*)\\\s/.freeze
+      LIST_BULLETS_OPTIONAL = '(\d{1,2}.|\*)\s*'
```
forelabs, line 14: I think this should be optional here.
```suggestion
      LIST_BULLETS_OPTIONAL = '(\d{1,2}.|\*)?\s*'
```

lib/license_finder/license/text.rb
```diff
@@ -6,15 +6,28 @@ module Text
...
 
       def self.normalize_punctuation(text)
-        text.gsub(SPACES, ' ')
-            .gsub(QUOTES, '"')
-            .strip
+        text.dup.force_encoding('UTF-8')
+                 .gsub(SPECIAL_DOUBLE_QUOTES, '"')
```
forelabs, line 18: I also found some licenses using quote characters or comment characters and also escape quotes already.
What do you think additional about
```suggestion
                 .gsub(/\>+/, '')
                 .gsub(/\*+/, '')
                 .gsub(/\\\"/, '"')
                 .gsub(SPECIAL_DOUBLE_QUOTES, '"')
```

lib/license_finder/license/text.rb
```diff
@@ -6,15 +6,28 @@ module Text
...
+                 .gsub(SPECIAL_SINGLE_QUOTES, "'")
+                 .gsub(SPACES, ' ')
+                 .gsub(QUOTES, '"')
+                 .strip
```
forelabs, line 22: some people also use different line breaks for a license, I think this should also be no issue to just ignore them, what do you think?

```suggestion
                 .gsub(/\n+/, ' ')
                 .strip
```

lib/license_finder/license/text.rb
```diff
@@ -6,15 +6,28 @@ module Text
...
       end
 
       def self.compile_to_regex(text)
-        Regexp.new(Regexp.escape(text).gsub(PLACEHOLDERS, '(.*)'))
+        Regexp.new(Regexp.escape(normalize_punctuation(text))
+                       .gsub(PLACEHOLDERS, '(.*)')
+                       .gsub(',', '(,)?')
```
forelabs, line 28: same here with new lines, we could either remove them or keep them optional.

```suggestion
                       .gsub(/\n+/, ' ')
                       .gsub(',', '(,)?')
```

forelabs:
I hope it is fine to review this - we are currently in the process of migrating to this tool and also need better detection here.

lib/license_finder/license/text.rb
forelabs: I have also recognized that some people have changed `COPYRIGHT HOLDER` to `COPYRIGHT OWNER`.

```ruby
.gsub('HOLDER', '(HOLDER|OWNER)')
```

lib/license_finder/license/text.rb
```diff
@@ -6,15 +6,37 @@ module Text
...
+      ALPHABET_ORDERED_LIST_OPTIONAL = '(\([a-z]\)\s)?'
+      LIST_BULLETS = /(\d{1,2}\\\.|\\\*)\\\s/.freeze
+      LIST_BULLETS_OPTIONAL = '(\d{1,2}.|\*)?\s*'
+      NEWLINE_CHARACTER = /\n+/.freeze
+      QUOTE_COMMENT_CHARACTER = /^\s*\>+/.freeze
+      ESCAPED_QUOTES = /\\\"/.freeze
 
       def self.normalize_punctuation(text)
-        text.gsub(SPACES, ' ')
+        text.dup.force_encoding('UTF-8')
```
forelabs, line 20: force encoding does not change the string data but just makes sure that e.g. raw data is read as utf-8. But in our cases we have some text with invalid bytes for utf-8 so this will fail. I need to use

```suggestion
        text.dup.encode('UTF-8', invalid: :replace, undef: :replace, replace: '?')
```

would this also be a possible general solution?

xtreme-debbie-chen: Unfortunately, `.encode` is insufficient in some cases. Specifically, I'm trying to add handling for smart quotes. I've been running my tests against this specific license: https://github.com/gregjones/httpcache/blob/master/LICENSE.txt

I did a quick test where I dropped some non-UTF-8 characters into that license and the `force_encoding` seems to handle it nicely. Is there a specific license that you're looking at where it fails?

lib/license_finder/license/text.rb
```diff
@@ -6,15 +6,37 @@ module Text
...
+            .gsub(SPECIAL_DOUBLE_QUOTES, '"')
+            .gsub(SPECIAL_SINGLE_QUOTES, "'")
+            .gsub(QUOTE_COMMENT_CHARACTER, '')
+            .gsub(SPACES, ' ')
+            .gsub(NEWLINE_CHARACTER, ' ')
+            .gsub(ESCAPED_QUOTES, '"')
             .gsub(QUOTES, '"')
             .strip
+      rescue ArgumentError => _e
+        text
```
Pivotal-Jeff-Jun, line 30: If we fail to `force_encode` are we not going to try `gsub`?

lib/license_finder/license/text.rb
forelabs: Don't worry, I've found a way that it works with this one. I have an additional `RailsVendor` package manager (to detect `RailsVendorPackage`) because I am to lazy to add those dependencies as "hidden" ones. For that I parse the vendor file itself for package information like version and license. Just have to encode the file as utf-8 itself when reading it.

PR #724, opened 2020-05-11
NPM: Detect by package-lock.json

forelabs:
Since npm has envolved over the past and there should
always be a package-lock.json. when using npm we should
check for that file instead since package.json is also
available when using yarn.

A reason for that change is that when you are using yarn, it will detect both NPM and Yarn but it should only detect one of both or if you have - for whatever reason - both a `yarn.lock`  and a `package-lock.json` then it will still detect both.

If the simple detection is needed and this won't get merged I would suggest at least an additional test that `yarn.lock` is not present as long as you don't see a reason why there should be both at the same time required.

### DIFF
--- a/features/support/testing_dsl.rb
+++ b/features/support/testing_dsl.rb
@@ -148,6 +148,7 @@ module LicenseFinder
 
     class NpmProject < Project
       def add_dep
+        add_to_file('package-lock.json', '')
         add_to_file('package.json', '{"dependencies" : {"http-server": "0.11.1"}}')
       end
 
@@ -158,6 +159,7 @@ module LicenseFinder
 
     class NpmProjectWithInvalidDependency < Project
       def add_dep
+        add_to_file('package-lock.json', '')
         add_to_file('package.json', '{"dependencies" : {"gertie-watch": "0.11.1"}}')
       end
 
--- a/lib/license_finder/package_managers/npm.rb
+++ b/lib/license_finder/package_managers/npm.rb
@@ -17,6 +17,10 @@ module LicenseFinder
       'npm install --no-save'
     end
 
+    def active?
+      project_path.join('package-lock.json').exist?
+    end
+
     def possible_package_paths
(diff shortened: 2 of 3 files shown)


xtreme-shane-lattanzio:
@forelabs Thank you so much for your contributions today! We are always open for improvements to LF as well as any input from the community on existing PRs or code. As you saw, we are also in the middle of improving our matching and having input from the community is very valuable due to having more use cases in the wild. We were actually recently talking about the proper way to get install paths for yarn as well as properly detecting Go Workspace projects so this helps us out quite a bit!

I have merged all of your PRs except this one as it has tests failing in CI. Also I am wondering if we want to do this because I worry about a case where someone does not commit their `package.lock` file. That would mean the LF would not trigger for that project. I'm open to justifications but just something that came to mind. Thanks again :)

forelabs:
Thank you, I have some more features introduced in our sub tool which I want to provide step by step here (e.g. package url generation like ruby gems, yarn.pm,... or copyright detection like copyright by author and others). 

Regarding package-lock.json:
That was why I also had some concerns about it and already suggested a different solution. But maybe a more general solution could be introduced. How about the possibility to deactivate package manager detections? Eg introduce a flag `ignored-package-managers` where someone could list ignored package managers by name or as a subcommand with list, add, remove and we store those decisions in the decision file (same behavior as for ignored dependencies). By using this, one could have npm and yarn or go modules and go vendor activated while another could simply ignore one of both if one is enough.

xtreme-shane-lattanzio:
Having ignored package manager options is a very interesting thought. Right now we have a way to ignore some package managers with `takes_priority_over` but it makes some assumptions about what kind of code a user may have. It works okay for Go cases but Yarn vs NPM is not as obvious. I would be open to an optional ignored package manager flow.

forelabs:
I will close this one and work on a `ignored_package_managers` command soon. I think it will be a better solution which might also be useful for other cases.

PR #727, opened 2020-05-12
Decision Applier: Merge manual and system packages

forelabs:
This will enable merging manual added dependencies with
detected dependencies. This is usefull if a package will
be detected but e.g. license information is missing.
This change will use data from decision (homepage, approval, license)
and apply it to system package matching the manual one.
Currently manual decisions override the whole system package
which results in losing information (e.g. package_manager)

### DIFF
--- a/lib/license_finder/decision_applier.rb
+++ b/lib/license_finder/decision_applier.rb
@@ -4,7 +4,7 @@ module LicenseFinder
   class DecisionApplier
     def initialize(options)
       @decisions = options.fetch(:decisions)
-      @all_packages = decisions.packages + options.fetch(:packages)
+      @all_packages = options.fetch(:packages).to_set + @decisions.packages.to_set
       @acknowledged = apply_decisions
     end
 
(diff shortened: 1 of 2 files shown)


lib/license_finder/decision_applier.rb
```diff
@@ -28,10 +28,14 @@ def any_packages?
...
-        .map { |package| with_approval(package) }
-        .map { |package| with_homepage(package) }
         .reject { |package| ignored?(package) }
+        .map do |package|
+          with_homepage(
+            with_approval(
+              with_decided_licenses(package)
+            )
+          )
+        end
```
forelabs, line 38: I thought we can reduce iterations, within one of our projects we have thousands of packages and reducing number of iterations can be already recognized regarding runtime.

spec/lib/license_finder/decision_applier_spec.rb
```diff
@@ -14,6 +14,15 @@ module LicenseFinder
 
     describe '#acknowledged' do
       it 'combines manual and system packages' do
+        decision_applier = described_class.new(
+          decisions: Decisions.new.add_package('system', nil).license('system', 'MIT'),
+          packages: [Package.new('system', '1.0.0')]
+        )
+        package = decision_applier.acknowledged.first
+        expect([package.name, package.version, package.licenses.first.name]).to match_array %w[system 1.0.0 MIT]
```
forelabs, line 22: The current master implementation would result in `system nil MIT`

PR #729, opened 2020-05-12
Package: add package url

forelabs:
I would like to introduce `package_url` to packages which will hold the url to the package details.
I tried to introduce it for every package manager, could not find a package registry for carthage and stb packages.

In a second step I would like to enable csv export of the `package_url` as a new column. An additional PR for this will follow when this one will be merged.

### DIFF
--- a/lib/license_finder/package.rb
+++ b/lib/license_finder/package.rb
@@ -43,6 +43,7 @@ module LicenseFinder
       @summary = options[:summary] || ''
       @description = options[:description] || ''
       @homepage = options[:homepage] || ''
+      @package_url = options[:package_url] || ''
       @children = options[:children] || []
       @parents = Set.new # will be figured out later by package manager
       @groups = options[:groups] || []
@@ -61,7 +62,7 @@ module LicenseFinder
 
     ## DESCRIPTION
 
-    attr_accessor :homepage
+    attr_accessor :homepage, :package_url
 
     attr_reader :name, :version, :authors,
                 :summary, :description,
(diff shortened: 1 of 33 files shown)


xtreme-shane-lattanzio:
@forelabs Thanks again for your PR! I am curious if the `homepage` field satisfies this criteria. I know that that field comes from what is found in the package manager itself and I am just wondering if these two fields would be used differently but users or if they are just different ways to get the same information?

forelabs:
@xtreme-shane-lattanzio I think those are different. E.g. for jquery homepage is `https://jquery.com/` but it is not that easy to get all relevant information from this site like verify license, see authors, ...
To provide an url where one could fetch easily this data from we need to provide the package url.
If this is special for us I am fine not getting this into the gem but I would guess that this information might be relevant for some people - at least a few of our customers would like to get this information to verify our fetched data.

pivotal-pmital:
@forelabs Is it intentional that the `package_url` field for `Bower` and `yarn` aren't version specific? We noticed that all the other package managers are but these 2 were not.

forelabs:
> @forelabs Is it intentional that the `package_url` field for `Bower` and `yarn` aren't version specific? We noticed that all the other package managers are but these 2 were not. 

I would like to have version specific urls for those too but actually the registries don't provide that possibility. The versions are visible as lists on those sites.

pivotal-pmital:
@forelabs I found [this](repo) where they mention a curl command -  
```
curl https://registry.bower.io/packages/<package_name>
```

This is providing a json response with a proper URL to the package info. Do you think this would be a better solution to your use-case and also a better alternative to the current `search` api call to Bower?

forelabs:
@pivotal-pmital Thank you for that, since bower relays on github packages only I think this is a good way. I have changed implementation to fetch url from the registry endpoint.

Issue #702, opened 2020-03-31
lookup dependency_decisions.yml from an url


micheelengronne:
If `LicenseFinder` can look at the `dependency_decisions.yml` file from an url, it will be a centralized way to manage it.

forelabs:
@micheelengronne maybe this will fit your needs https://github.com/pivotal/LicenseFinder/pull/735

Issue #587, opened 2019-05-24
Multiple decisionsfile location/inherit decisionsfile from other file


DarshitaPadaria:
Is there any way in which I can provide multiple decision files? Or any way in which I can inherit the decisions file from a parent decisions file.

I need a master decisions file which can be included/inherited into several project specific decisions file.

xtreme-shane-lattanzio:
@DarshitaPadaria We currently only support one decision file. If someone wants to PR a way to support an array of files or something like that, we could be open to it.

DarshitaPadaria:
@xtreme-shane-lattanzio , It would be great if support for multiple decision files are added. We have a hierarchy of maven projects. We need to update each project if there is something changed in parent project. Multiple files would help to manage these easily.

forelabs:
@DarshitaPadaria maybe this will fit your needs https://github.com/pivotal/LicenseFinder/pull/735

PR #735, opened 2020-05-18
Feature: Decision inheritance

forelabs:
## Decision Inheritance

- Add possibility to inherite from other decision files (solves https://github.com/pivotal/LicenseFinder/issues/587)
- Lookup decision from url (kinda solves https://github.com/pivotal/LicenseFinder/issues/702 by adding this for the inheritance feature)

Introducing new `inherited_decisions` command

```
license_finder inherited_decisions [list|add|remove]   # Add or remove decision files you want to inherit from - see `license_finder inherited_decisions help` for more information
```

```
Commands:
  license_finder inherited_decisions add DECISION_FILE...     # Add one or more decision files to the inherited decisions
  license_finder inherited_decisions list                     # List all the inherited decision files
  license_finder inherited_decisions remove DECISION_FILE...  # Remove one or more decision files from the inherited decisions
```

By providing this feature you could e.g. have a centralized decision file for approved/restricted licenses. If you have multiple projects it's way easier to have one single place where you approved or restricted licenses defined.

### DIFF
--- a/lib/license_finder/decisions.rb
+++ b/lib/license_finder/decisions.rb
@@ -186,10 +186,7 @@ module LicenseFinder
     def inherit_from(filepath)
       decisions =
         if filepath =~ %r{^https?://}
-          # ruby < 2.5.0 URI.open is private
-          # rubocop:disable Security/Open
-          open(filepath).read
-          # rubocop:enable Security/Open
+          open_uri(filepath).read
         else
           Pathname(filepath).read
         end
@@ -214,6 +211,17 @@ module LicenseFinder
       self.class.restore(decisions, self)
       @inherited = false
       self
+    end
+
+    def open_uri(uri)
+      # ruby < 2.5.0 URI.open is private
+      if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.5.0')
+        # rubocop:disable Security/Open
+        open(uri)
+        # rubocop:enable Security/Open
+      else
+        URI.open(uri)
+      end
     end
 
     #########
--- a/lib/license_finder/decisions.rb
+++ b/lib/license_finder/decisions.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'open-uri'
+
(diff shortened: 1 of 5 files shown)


lib/license_finder/decisions.rb
```diff
@@ -115,71 +118,104 @@ def approve(name, txn = {})
...
       self
     end
 
+    def inherit_from(filepath)
+      decisions =
+        if filepath =~ %r{^https?://}
+          # ruby < 2.5.0 URI.open is private
+          # rubocop:disable Security/Open
+          open(filepath).read
+          # rubocop:enable Security/Open
```
forelabs, line 192: I already used `open` in the `package_url` PR but rubocop did not complained about it. Here it does but for ruby 2.3.8 and ruby 2.4.9 it fails by using `URI.open` since for those versions this method is private. What do you suggest? Disable rubocop here or having a if/else block where I check the ruby version?

forelabs:
**concourse-ci/ruby-2.7.0** check is currently not running

xtreme-shane-lattanzio:
@forelabs This looks good! Thanks again!! re the concourse check, we updated the pipeline to look at 2.7.1 but forgot the git hook. All fixed now :)

lib/license_finder/decisions.rb
xtreme-shane-lattanzio: Hmm....good catch. I wonder if it was fine before because if the nested nature of how it was used so rubocop doesn't notice it. If URI.open is not available on 2.3 or 2.4, we may need to disable rubocop here for those versions as you have done. I would add the conditional for the versions though. Rubocop in the pipeline does currently run on the 2.7.1 ruby version of the ruby docker image so there shouldn't be an issue.

forelabs: I have added the version conditional as requested but kept the rubocop disable for the used `open` call for ruby < 2.5.0 otherwise rubocop would fail on development systems which I think should not happen.

PR #753, opened 2020-06-19
Update Nuget Package Manager prepare command

xtreme-jason-smith:
Nuget failed to 'restore' projects when it only contained the
'packages.config' file. Nuget would complain that it did not know where
to download the packages to.

So, the Nuget package manager has been updated to try the original
command, and if that fails, it would try again but this time providing
an -PackagesDirectory which is set to the current directory.

[#173420224]

Signed-off-by: Jason Smith <jsmith@pivotal.io>

### DIFF
--- a/lib/license_finder/package_managers/nuget.rb
+++ b/lib/license_finder/package_managers/nuget.rb
@@ -1,4 +1,4 @@
-# frozen_string_literal: true
+# frozen_string_literal: false
 
 require 'rexml/document'
 require 'zip'
@@ -87,6 +87,23 @@ module LicenseFinder
       return 'nuget' if LicenseFinder::Platform.windows?
 
       "mono #{nuget_binary}"
+    end
+
+    def prepare
+      _, stderr, status = Dir.chdir(project_path) { Cmd.run(prepare_command) }
+      return if status.success?
+      log_errors stderr
+      logger.info prepare_command, 'trying fallback prepare command', color: :magenta
+
+      cmd = "#{prepare_command} -PackagesDirectory ."
+      stdout, stderr, status = Dir.chdir(project_path) { Cmd.run(cmd) }
+      return if status.success?
+      log_errors_with_cmd(cmd,stderr)
+
+      error_message = "Prepare command '#{cmd}' failed\n#{stderr}"
+      error_message = error_message.concat("\n#{stdout}\n") if !stdout.nil? && !stdout.empty?
+
+      raise error_message unless @prepare_no_fail
     end
 
     def prepare_command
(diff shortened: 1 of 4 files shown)


lib/license_finder/package_managers/nuget.rb
```diff
@@ -89,6 +89,23 @@ def package_management_command
...
+      log_errors stderr
+      logger.info prepare_command, 'trying fallback prepare command', color: :magenta
+
+      cmd = "#{prepare_command} -PackagesDirectory ."
+      stdout, stderr, status = Dir.chdir(project_path) { Cmd.run(cmd) }
+      return if status.success?
+      log_errors_with_cmd(cmd,stderr)
+
+      error_message = "Prepare command '#{cmd}' failed\n#{stderr}"
+      error_message = error_message.concat("\n#{stdout}\n") if !stdout.nil? && !stdout.empty?
```
Manifaust, line 104: You can try using `+` if you want to retain frozen_string_literal and see if that works

```suggestion
      error_message = error_message + "\n#{stdout}\n" if !stdout.nil? && !stdout.empty?
```

lib/license_finder/package_managers/nuget.rb
```diff
@@ -89,6 +89,23 @@ def package_management_command
       "mono #{nuget_binary}"
     end
 
+    def prepare
+      _, stderr, status = Dir.chdir(project_path) { Cmd.run(prepare_command) }
+      return if status.success?
+      log_errors stderr
+      logger.info prepare_command, 'trying fallback prepare command', color: :magenta
+
+      cmd = "#{prepare_command} -PackagesDirectory ."
```
Manifaust, line 98: It would be nice if you could check the error message for the exact thing you're trying to fix:
`Please specify either -PackagesDirectory or -SolutionDirectory.` before executing this fallback. Unless you think this fallback is a useful general solution.

xtreme-jason-smith: Yes, we could. My concern is that if they change that string, we may not notice and we will still use the fallback.  But maybe we could look for keywords, such as `-PackagesDirectory` or something maybe?

lib/license_finder/package_managers/nuget.rb
xtreme-jason-smith: Yes, that sounds like a good idea. We can give it a shot on Monday.

lib/license_finder/package_managers/nuget.rb
Manifaust: Yeah that sounds good

PR #763, opened 2020-07-07
[Added] support for rebar3

Pivotal-Jeff-Jun:
[Removed] support for rebar2

Rebar2 seems to be last updated in 2016, and seems to be not actively supported, so we decided to bump our rebar version to rebar3.

[#173637980]

### DIFF
--- a/lib/license_finder/package_managers/rebar.rb
+++ b/lib/license_finder/package_managers/rebar.rb
@@ -10,7 +10,7 @@ module LicenseFinder
 
     def current_packages
       rebar_deps.map do |name, version|
-        licenses, homepage = rebar_pkgs(name)
+        licenses, homepage = dep_info(name)
         RebarPackage.new(
           name,
           version,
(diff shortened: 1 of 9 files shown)


lib/license_finder/package_managers/rebar.rb
```diff
@@ -30,15 +32,34 @@ def possible_package_paths
...
-        .reject { |line| line.start_with?('=') }
-        .map { |line| line.split(' ') }
+        .reject { |line| line.start_with?('=') || line.include?('project app') }
+        .map do |line|
+          matches = line.match(/(?<name>\w+)─(?<version>[\S.]+)\s*/)
+          [matches[:name], matches[:version]] if matches
+        end.compact
+    end
+
+    def rebar_pkgs(name)
```
pivotal-pmital, line 49: Is it better to name the method after the CLI method that we're calling? Or is it better to name it in a way that tells what purpose it's serving for us?

Issue #654, opened 2020-01-06
Non-vendored Go modules are not properly detected


sgnn7:
When using this codebase against a `go.mod` `GO111MODULES`-backed codebase that doesn't have code vendored, no license data for dependencies is returned as the path prefix for the packages is marked as `<CWD>/vendored/...`. May be related to [this](https://github.com/golang/go/issues/35589) upstream issue when the listing is done [here](https://github.com/pivotal/LicenseFinder/blob/master/lib/license_finder/package_managers/go_modules.rb#L24)

To replicate save and run this file:
```bash
#!/bin/bash -e

docker run --rm \
           licensefinder/license_finder \
           /bin/bash -elc "mkdir /scan &&
                           cd /scan &&
                           wget -q https://raw.githubusercontent.com/cyberark/conjur-api-go/master/go.mod
                           wget -q https://raw.githubusercontent.com/cyberark/conjur-api-go/master/go.sum
                           license_finder report"
```

Output (notice the `unknown`s for everything):
```
$ ./runit.sh 
mesg: ttyname failed: Inappropriate ioctl for device
LicenseFinder::GoModules: is active
github.com/bgentry/go-netrc, v0.0.0-20140422174119-9fd32a8b3d3d, unknown
github.com/gopherjs/gopherjs, v0.0.0-20181103185306-d547d1d9531e, unknown
github.com/jtolds/gls, v4.2.2-0.20181110203027-b4936e06046b+incompatible, unknown
github.com/sirupsen/logrus, v1.0.5, unknown
github.com/smartystreets/assertions, v0.0.0-20190215210624-980c5ac6f3ac, unknown
github.com/smartystreets/goconvey, v0.0.0-20190222223459-a17d461953aa, unknown
golang.org/x/crypto, v0.0.0-20180621125126-a49355c7e3f8, unknown
golang.org/x/sys, v0.0.0-20180709060233-1b2967e3c290, unknown
gopkg.in/yaml.v2, v2.2.1, unknown
```

xtreme-shane-lattanzio:
@sgnn7 Those paths listed are the ones that are explicitly inside of the go.mod file. The license is unknown because there is no way to find the license unless we actually hit the web. Is this more about the fact they are unknown or you are expecting to see more than just the vendored stuff because of the `-mod=vendor` flag? This was done a while ago so I forget why we explicitly put that flag in there :S

sgnn7:
@xtreme-shane-lattanzio 

[...]

PR #764, opened 2020-07-07
Fix Go modules package manager

miquella:
While upgrading our Go project from Glide to Go modules, we noticed a number of discrepancies between the two implementations. This PR attempts to address these inconsistencies.

Each fix has been separated into its own commit for ease of review, but I am happy to separate these into individual PRs as well, if that makes things easier.

The issues discovered are as follows:

1. The `go.mod` file should be used to identify a module, not `go.sum` (667f6be716504a53ccc2824daae08af085566546)
2. The `go.mod` file is discovered based on the current working directory, so the working directory needs to be changed to the project path before invoking `go` (667f6be716504a53ccc2824daae08af085566546)
3. The `go list` command requires no preparation, tidying, nor vendoring (284cc5c821270a6e56275e32bac836a3e451f46b)
4. Only imported packages/modules should be listed (just like other Go package manager implementations) (34361fdab2dc3f197f7aec6408175018dee3b453 and dffae4ab95e34115b6a54bf681fc0966a8611f01)

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ languages, as long as that language has a package definition in the project dire
 * `glide.lock` file (for `glide`)
 * `vendor/vendor.json` file (for `govendor`)
 * `Gopkg.lock` file (for `dep`)
-* `go.sum` file (for `go mod`)
+* `go.mod` file (for `go mod`)
 * `vendor.conf` file (for `trash`)
 * `yarn.lock` file (for `yarn`)
 * `conanfile.txt` file (for `conan`)
--- a/lib/license_finder/package_managers/go_modules.rb
+++ b/lib/license_finder/package_managers/go_modules.rb
@@ -30,12 +30,30 @@ module LicenseFinder
 
     def packages_info
       Dir.chdir(project_path) do
-        info_output, stderr, _status = Cmd.run("GO111MODULE=on go list -m -f '{{.Path}},{{.Version}},{{.Dir}}' all")
-        if stderr =~ Regexp.compile("can't compute 'all' using the vendor directory")
-          info_output, _stderr, _status = Cmd.run("GO111MODULE=on go list -m -mod=mod -f '{{.Path}},{{.Version}},{{.Dir}}' all")
-        end
+        # Explanations:
+        # * Ignore standard library packages
+        #   (not .Standard)
+        # * Replacement modules are respected
+        #   (or .Module.Replace .Module)
+        # * Module cache directory or (vendored) package directory
+        #   (or $mod.Dir .Dir)
+        format_str = \
+          '{{ if not .Standard }}'\
+            '{{ $mod := (or .Module.Replace .Module) }}'\
+            '{{ $mod.Path }},{{ $mod.Version }},{{ or $mod.Dir .Dir }}'\
+          '{{ end }}'
 
-        info_output.split("\n")
+        # The module list flag (`-m`) is intentionally not used here. If the module
+        # dependency tree were followed, transitive dependencies that are never imported
+        # may be included.
+        #
+        # Instead, the owning module is listed for each imported package. This better
+        # matches the implementation of other Go package managers.
+        info_output, stderr, _status = Cmd.run("GO111MODULE=on go list -f '#{format_str}' all")
+        info_output, _stderr, _status = Cmd.run("GO111MODULE=on go list -mod=mod -f '#{format_str}' all") if stderr =~ Regexp.compile("can't compute 'all' using the vendor directory")
+
+        # Since many packages may belong to a single module, #uniq is used to deduplicate
+        info_output.split("\n").uniq
       end
     end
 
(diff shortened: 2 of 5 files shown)


pivotal-issuemaster:
@miquella Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=764)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=764) to manually synchronize the status of this Pull Request.

miquella:
This looks like it may help address some of the issues with #654 as well.

pivotal-issuemaster:
@miquella Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=764)!

miquella:
I've realized one additional tweak that needs to be made. I'll test it and get it pushed up tomorrow morning.

miquella:
Apparently I'm going to fix this instead of sleep :stuck_out_tongue_closed_eyes:

From all of my testing, this finally appears to be well in line with the other Go package manager implementations and is working well for us.

xtreme-shane-lattanzio:
Hi @miquella ! Thanks so much for this PR. It is great to have someone with deeper go insight take a look at this flow. I was surprised to see that to tidy commands are not needed for this flow. 

This looks good to merge in. I am just wondering about that last TODO that you put in the comments. Is that for the future as this is currently functional or would you like use to wait for that implementation?

miquella:
I'm glad I had some time to tackle this!

----

Regarding the prepare commands:

* The `go mod vendor` command is no longer useful as the vendor directory is now explicitly ignored (`-mod=readonly` disables reading the vendor directory)
* The `go mod tidy` command was likely useful when using `go list` in module mode (`-m`), but with the shift to tracing imported packages, it's not as useful (a lot of what tidy does is remove unused dependencies, which will be implicitly excluded when tracing imported packages)

----

As for the TODO: yes, it was meant as a future thing to look into. I tried poking at it a bit with vendor mode enabled (`-mod=vendor`), but it looks like it'll require traversing up through the directory tree to locate the root of each module. It certainly sounds doable, but will be a bit more work.

My first thought is to iterate through each of the listed package directories, `Dir.chdir` into the directory, and use something like `go env GOMOD` to find the root directory of the module. This would allow us to coalesce all related packages into a single module description.

This actually doesn't sound too bad, so maybe I'll give it a quick shot and get back to you if it turns into something too large for me to tackle right now :smiley:

miquella:
@xtreme-shane-lattanzio: Well, unfortunately my first thought doesn't work :disappointed:

I adjusted `go list` to output both the module path as well as the package path. Then if the module path was empty, I tried to determine what it was based on the package path. Unfortunately, `go mod vendor` is excluding some `go.mod` files while including others, so `go env GOMOD` can't find the module's root.

This is extra strange since the module cache (which is what the code in the current PR uses) _does_ include the `go.mod` files…?

Perhaps someone will come up with something clever one day, but I don't have any other ideas right now.

xtreme-shane-lattanzio:
@miquella No problem! Just one more question more for my curiosity. On older gomod versions`-mod=mod` was default. When we upgraded here https://github.com/pivotal/LicenseFinder/commit/0969e98fde4a82f8931601baa4dd96dc01300a14#diff-7291a2df7ea62cdb37ebeafccb69088f we had to flip it because `-mod=vendor` became default. We used to do a check that said try to use vendor but if it fails, use mod. Here you are using readonly. Does that prevent the need for this check? I know you said it will disable the vendor directory so maybe thats why this works?

Please forgive the fact that I am not clear on what either of the flags actually do :P

miquella:
@xtreme-shane-lattanzio: No problem at all. Happy to answer any questions I can!

There are a couple of things all balled up together here. First, Go 1.14 changed the default for `-mod` to be `vendor` when a vendor directory is present. That same version also introduced the `mod` option for this flag to disable this default behavior around vendoring (the `mod` option didn't exist before 1.14).

Related to this, the `readonly` option of the `-mod` flag does the same thing as the new `mod` option, but disables modification of the `go.mod` file. But the `readonly` option was introduced with Go modules in version 1.11, unlike the `mod` option.

But I believe the core of your question is around the fallback condition when the "can't compute 'all' using the vendor directory" error is reported. In doing some research around that, I discovered that it's an error specific to `go list` being used in module mode (`-m`), which we're no longer using :smile: (which is also evident in the spec, which mocks the error as: `go list -m: …`)

xtreme-shane-lattanzio:
Sounds good. Makes sense. Thanks for the explanation! Will merge in :D

miquella:
@xtreme-shane-lattanzio: Oh, I don't think I understood how to structure my messages so they would end up in the CHANGELOG. Would you like me to push a PR that manually adds a few notes to the CHANGELOG?

xtreme-shane-lattanzio:
@miquella Sounds good to me. Thanks for doing that :)

Issue #291, opened 2017-03-21
Please add conda environment support


emiola:
It would be nice to support [conda](https://github.com/conda/conda) [`environment.yml` files](https://github.com/conda/conda/tree/25e52acaf567e30e73fe43cbadb02f5b6a8ab6d7/conda_env#environmentyml).

dpasqualin:
I see this issue was opened a long time ago. Is there any perspective on the implementation of this feature?

kdykeman:
If somebody more experienced with Conda could put together a PR, we'd definitely look at pulling it in.

PR #776, opened 2020-08-20
Added support for Conda (experimental, focused on Python)

cobbr2:
Addresses https://github.com/pivotal/LicenseFinder/issues/291

Notes:
* Conda is a shell-based environment like `rvm` or `asdf`, so takes some special handling of the command line
* Assumes `bash` is available in the path
* Made no effort to support Windows, do not have test environment where I could
* Tests and initial operations have been focused on Python-based projects deployed in Linux environments
* Has not been tested outside of `dlf` containers
* Apologies for removing line-ending spaces in `.md` files; my editor is fastidious in that way
* Can not claim to be a `conda` expert, but this works for our projects


pivotal-issuemaster:
@cobbr2 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=776)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=776) to manually synchronize the status of this Pull Request.

lib/license_finder/package_managers/conda.rb
```diff
@@ -0,0 +1,132 @@
+# frozen_string_literal: true
+
+require 'json'
+require 'pry'
```
xtreme-shane-lattanzio, line 4: Remove require pry

lib/license_finder/package_managers/conda.rb
```diff
@@ -0,0 +1,132 @@
...
+
+module LicenseFinder
+  class Conda < PackageManager
+    attr_reader :conda_bash_setup_script
+
+    def initialize(options = {})
+      @conda_bash_setup_script = options[:conda_bash_setup_script] || Pathname('/root/miniconda3/etc/profile.d/conda.sh')
```
xtreme-shane-lattanzio, line 11: Asking more for curiosity. Is hardcoding this path the right thing to do? Does this script normally sit here? I do like how you made this customizable

Dockerfile
```diff
@@ -159,11 +159,21 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 4F4EA0AAE5
     apt-get update &&\
     apt-get install -y php7.4-cli &&\
     php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" &&\
-    php -r "if (hash_file('sha384', 'composer-setup.php') === 'e5325b19b381bfd88ce90a5ddb7823406b2a38cff6bb704b0acc289a09c8128d4a8ce2bbafcd1fcbdc38666422fe2806') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;" &&\
+    php -r "if (hash_file('sha384', 'composer-setup.php') === '572cb359b56ad9ae52f9c23d29d4b19a040af10d6635642e646a7caa7b96de717ce683bd797a92ce99e5929cc51e7d5f') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;" &&\
```
xtreme-shane-lattanzio, line 162: This was changed again on master so please rebase to get the new sha

features/support/testing_dsl.rb
```diff
@@ -112,6 +112,33 @@ def clone(fixture_name)
     class EmptyProject < Project
     end
 
+    class CondaProject < Project
+      CONDA_ROOT = '/root/miniconda3'
```
xtreme-shane-lattanzio, line 116: Same question as before about hardcoded paths. Not a problem, just curious if this will always be the case or just most of the time.

lib/license_finder/package_managers/conda.rb
cobbr2: Thanks for asking. This is a reasonable default since it matches up to where it's installed in the Dockerfile, but there's no predictable "normal" place (other than `/etc/profile.d`, but many [all of our] developers install to `~/miniconda3`).  OTOH, I think there's a way to simplify this handling and will make an attempt to do so.

cobbr2:
Still waiting on our legal team to approve the CTO signing the CLA.  Probably will be a week before that happens.

lib/license_finder/package_managers/conda.rb
cobbr2: Unfortunately, the best I could do was make the default be based on $HOME instead of hardcoding `/root/`.

cobbr2:
Our CTO, Wade Chambers, has signed the CLA, and gotten an indication that the block on this PR should now clear. I presume there's some batch thing that'll make that happen, and will check in later to see if there's anything else I should be doing.

Pivotal-Jeff-Jun:
@cobbr2 Looks like the person who opened the PR needs to sign the CLA, so could you sign the CLA form?
Thanks. 😄

features/support/testing_dsl.rb
cobbr2: It will always be the same since it's driven by the way we install conda in the container.

pivotal-issuemaster:
@cobbr2 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=776)!

xtreme-shane-lattanzio:
Hey @cobbr2 ! Thanks for signing the CLA. It looks like the key for Miniconda3 is now incorrect when building since its been a while. I wonder if there is a way to fetch the right sha so this does not fail. Maybe something similar to the way that Composer is installed in the Dockerfile?

cobbr2:
Looks like reference sha has changed again. One more commit.

cobbr2:
I'll see what the composer one does.  Thanks for the pointer.

cobbr2:
Just went ahead and pinned to the latest current version and its sha; amusingly, I hadn't pulled master, so hadn't seen the updates to the way Composer was installed.  I didn't see a similar set of resources in the miniconda's repos (not to say they aren't there, just not obvious), so pinning seemed like a good alternative.

xtreme-shane-lattanzio:
@cobbr2 Sure we can keep pinning for now as long as it does not change often. However it is still failing with your current changes. Can you take a look?

cobbr2:
Thanks for the merge!

PR #781, opened 2020-08-31
Added `git submodule` "package manager"

### LINKED ISSUE #780: git submodule support
Some project don't use a specific package manager, but depend on `git submodule` (older projects, C, C++,... (c)make based stuff...).

It would be nice if this was added to licenseFinder (`git submodule status` and the content of `.gitmodules` files have the needed info). Package name would be a bit harder (either the git path, or the path where it is checked out), but with some liberty this would work out fine.

wonko:
- parses the git submodule status output for versions
- parses the .gitmodules files for path and url info
- recurses into the submodules
- Added tests

Closes #780 

Example output:

```
LicenseFinder::GitSubmodule: is active
submodules/Fast-BVH, heads/master, MIT
submodules/apis, v3.11.0-rc.27-4-g8d20cb02, unknown
submodules/build-tools, heads/master, unknown
...
submodules/pybind11, v2.4.3-35-g4f72ef8, "New BSD"
tools/clang, 6a00cbc, unknown
```


pivotal-issuemaster:
@wonko Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=781)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=781) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@wonko Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=781)!

xtreme-shane-lattanzio:
Hi @wonko ! Thank you for the PR! This is an interesting one. Submodules is not exactly a package manager and I think this may actually cause some issues for users as it is now.

I know there are teams out there that use pipelines to manually pull in submodules and then run whatever they need to on the code. The submodules are treated differently that the parents in their flows.

I am wondering if a different approach would be to use this code but not to trigger on package manager type things. Perhaps a command line argument that people can use optionally to pull in submodules if found? What do you think?

I do think it is a cool idea to automatically pull in submodules but am a bit worried about backwards compatibility on this one. Thanks again!

wonko:
You're right as it might confuse people. Now, i think it might be a good idea to allow _all_ managers to be enabled/disabled per config and/or command line, so someone can be selective on what they want (we actually had a similar issue, LF detected our python env which implements some simple controls for our CI flow, and not so much the libs for the code we write and build). Then you could enable everything by default, or disable everything, or build in a bias (LF could even output that it detected a certain manager, but that it is disabled by default). Many options, but it somehow is bigger than this PR...

xtreme-shane-lattanzio:
The reason why I call out this PR specifically for optional trigger is that this will trigger on a very large set of git projects while using programming language specific triggers would be relatively smaller. There are ways around not triggering right now which are a big hacky such as a prepare step to remove the triggering file. We also have a method in the code `takes_priority_over` when you have two similar package managers but want one to always run instead of the other. That is primarily for Go cases right now.

But youre right. Im now thinking of some kind of config file to override/exclude certain package managers but again, it is beyond the scope of this PR lol. I think I am more worried about this specific one being enabled by default. Even if we implemented an exclude/allow feature, to ensure consistency, everything would be enabled by default which I think would be the opposite of submodules.

xtreme-shane-lattanzio:
Closing due to above concerns

PR #782, opened 2020-09-17
Fix godep not finding/downloading package sources [#174537978]

Pivotal-Jeff-Jun:
Signed-off-by: Julia Pu <jpu@pivotal.io>

### DIFF
--- a/spec/lib/license_finder/package_managers/go_dep_spec.rb
+++ b/spec/lib/license_finder/package_managers/go_dep_spec.rb
@@ -49,7 +49,7 @@ module LicenseFinder
         end
       end
 
-      context ' when there is a populated godep json' do
+      context 'when there is a populated godep json' do
         before do
           FakeFS.activate!
           FileUtils.mkdir_p '/fake/path/Godeps'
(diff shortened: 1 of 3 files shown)


ohlookadollar (approved):

spec/lib/license_finder/package_managers/go_dep_spec.rb
```diff
@@ -140,14 +140,10 @@ module LicenseFinder
           end
         end
 
-        context 'when dependencies are not vendored' do
```
xtreme-jason-smith, line 143: So, does this mean we no longer have a case where dependencies will not be vendered?

spec/lib/license_finder/package_managers/go_dep_spec.rb
```diff
@@ -156,10 +152,105 @@ module LicenseFinder
...
+              allow(FileTest).to receive(:directory?).with("/fake/path/#{GoDep::GODEP_VENDOR_PATH}").and_return(false)
+
+              expect(SharedHelpers::Cmd).to receive(:run).with('godep restore').and_return([nil, 'some-error-message', failed_status])
+            end
+
+            after do
+              ENV['GOPATH'] = @orig_gopath
+            end
+
+            it 'should raise error' do
```
xtreme-jason-smith, line 251: Is this going to be a hard stop? Meaning is LF suppose to quit if it cannot download the sources in this scenario?

xtreme-jason-smith (commented):
Looks good overall.  Just had one question.

spec/lib/license_finder/package_managers/go_dep_spec.rb
Pivotal-Jeff-Jun: The cases we think the download command would fail will be:
- the project is broken
- randomly just failed
- network died in the middle

In those cases, I think it would be better to error out instead of returning partial data.

Also, we were going through gradle, and looks like gradle also fails out when gradle command fails.
So we think it should be a hard stop, so that we can keep consistency between different package managers.

PR #796, opened 2020-11-13
Add --enabled-package-managers option to narrow down package managers

ganta:
I added support for specifying a package manager to be enabled.
This feature is intended for the following purposes:

- Switch the decisions file for each package manager
  (e.g. Separate licenses to be allowed on the server-side and front-end)
- Exclude specific package managers from the license check
  (e.g. Skip checking package managers that only resolve dependencies on development tools that do not affect the production code)


xtreme-shane-lattanzio:
@ganta Thanks for the PR! Looks good to me and I'm sure this will make people very happy :D

ganta:
@xtreme-shane-lattanzio 
Thanks for your review! I'm glad to have my PR merged😃

PR #801, opened 2020-12-08
Feature/spm packager

### LINKED ISSUE #676: Swift package manager support
It would be lovely if `licensefinder` could support Swift Package Manager.

https://github.com/apple/swift-package-manager

Dependencies are listed in a [Package.resolved](https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md#resolving-versions-packageresolved-file) file in the project with URLs to github or other hosted repos.

An example of [Package.resolved](https://github.com/thoughtbot/ArgoJSONAPI/blob/master/Package.resolved) file

michallaskowski:
This PR would close #676 .
There are two ways that SPM works - on Linux via `swift`, and on macOS it can be used via `xcodebuild`. Both cases are covered here, though for `xcodebuild` build path needs to be passed an env variable. 
In case of project-relative DerivedData, it would look like this:
```
SPM_DERIVED_DATA=./DerivedData/some-project/SourcePackages license_finder report
```


pivotal-issuemaster:
@michallaskowski Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=801)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=801) to manually synchronize the status of this Pull Request.

michallaskowski:
I am working on getting Corporate CLA signed. In the meantime, it seems feature tests for `sbt` failed. I swear I didn't touch it 🤞:)

xtreme-jason-smith:
> I am working on getting Corporate CLA signed. In the meantime, it seems feature tests for `sbt` failed. I swear I didn't touch it 🤞:)

Sounds good.

In the mean time we'll take a deeper look into the `sbt` failure.

Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM ubuntu:xenial
+# Ubuntu Xenial(16.04) with Swift 5.3.1 installed
+FROM swift:5.3.1-xenial
```
xtreme-shane-lattanzio, line 2: Is there another way to do this besides changing the base image? We probably do not want to change the entire image just to support a single package manager. There are also cases where other images are built off of LF and these cases would need to be tested by anyone that is doing that if were to make this change. It can be done but thinking there has to be a different way the required things in the swift image can be installed

xtreme-shane-lattanzio (changes requested):
Hey @michaelklishin ! Thanks for looking into this. This is something that has been asked for for a while. Only concern I have is about changing the base image. Please let me know your thoughts!

Dockerfile
michallaskowski: I am not an expert on Docker, so any suggestions welcome. It was the easiest way for me to try add support for SPM, because it was also based on Ubuntu 16.04. If someone can share what would be the best way to do it, I can give it a try soon.

michallaskowski: I can give it a try with https://docs.docker.com/develop/develop-images/multistage-build/

xtreme-shane-lattanzio: This is less of a docker question and more of a question of how install SPM. Looking [here](https://github.com/apple/swift-package-manager#installation), there are steps to install it and I am wondering if they will work. If you look in the Dockerfile, that is how other package managers are being installed. For testing purposes, what you can do is boot into the image then run the installation commands and check if it does install the swift cli properly. Once you know, you can add the steps to the docker image and then rebuild it locally to try it out

pivotal-issuemaster:
@michallaskowski Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal/LicenseFinder&pullRequestId=801)!

Dockerfile
michallaskowski: We could do that, it would basically be a copy paste from here: https://github.com/apple/swift-docker/blob/main/5.3/ubuntu/16.04/Dockerfile
Maybe to keep shorter gpg signature verification could be omitted. But those instructions for Linux would basically be it.
If copy-paste from there is ok, I can do it. 
Or could use a multistage build - and then do we want it is probably a Docker question. I am not aware of less obvious pros and cons, except that for future Swift versions it would be much easier to maintain multistage build, basically delegating the process of Swift installation to a Dockerfile maintained by Apple.

xtreme-shane-lattanzio: Copying and pasting is not a bad starting point for this. If you take a look at the LF dockerfile, we already have most package managers hard coded to a version at the top. The only issue I see happening is if the signing key changes and I don't know if there is an automated way to get it. If you look at the composer section, there is an automated way there to get keys

xtreme-shane-lattanzio:
Hey @michallaskowski ! Any plan on resuming work on this? If we can get the base image changed I think this could be useful!

michallaskowski:
@xtreme-shane-lattanzio Hey. Thanks for reminding, I was a bit occupied with other things. The only outstanding thing seems to be Docker setup, I just pushed changes. But for some reason I am having trouble running tests with `./dlf rake spec`.
Getting `Failure/Error: require 'xmlsimple'`
If my changes somehow made this happen, would be happy to get some help with fixing the Docker setup (still not an expert there ;))

xtreme-shane-lattanzio:
@michallaskowski Thanks for completing this! The jruby stuff is unrelated so no worries. There is a separate github issue filed for that [here](https://github.com/pivotal/LicenseFinder/issues/807)

mrkingfoxx (commented):
I am just learning sorry if I make a mistake

mrkingfoxx (commented):
Check

abinrm:
@michallaskowski Hi Michal, thanks for updating the repo with SPM support. How can we use this with xcodebuild? I added the environment variable `SPM_DERIVED_DATA` with correct folder location, what exactly should I enter in terminal to run license finder? This is what I get. Any leads would be helpful. Thank you.
```
abin@RM-1146 project-folder % license_finder --decisions_file licenses/licenses.yml                               
License Finder: No active and installed package managers found for project.

No dependencies recognized!
```

PR #816, opened 2021-02-22
Relax thor version requirement

nertzy:
I use license finder on some Ruby projects that also use thor.

I want to be able to upgrade to thor 1.1 on my project, but license finder has a strict 1.0.x version requirement. This relaxes the requirement to 1.x.

### DIFF
--- a/lib/license_finder/cli/approvals.rb
+++ b/lib/license_finder/cli/approvals.rb
@@ -13,7 +13,7 @@ module LicenseFinder
         assert_some names
         modifying { names.each { |name| decisions.approve(name, txn) } }
 
-        say "The #{names.join(', ')} dependency has been approved!", :green
+        say! "The #{names.join(', ')} dependency has been approved!", :green
       end
 
       auditable
@@ -21,7 +21,7 @@ module LicenseFinder
       def remove(dep)
         modifying { decisions.unapprove(dep, txn) }
 
-        say "The dependency #{dep} no longer has a manual approval"
+        say! "The dependency #{dep} no longer has a manual approval"
       end
     end
   end
--- a/lib/license_finder/cli/base.rb
+++ b/lib/license_finder/cli/base.rb
@@ -29,7 +29,7 @@ module LicenseFinder
       private
 
       def fail(message)
-        say(message) && exit(1)
+        say!(message) && exit(1)
       end
 
       def license_finder_config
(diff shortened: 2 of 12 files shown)


david-a-wheeler:
+1 on relaxing the thor version specification. I have multiple dependencies that depend on thor; this license_finder requirement forces my application backwards and in the future it may not be possible.

Good luck!

xtreme-shane-lattanzio:
Hey @nertzy ! I am all for this but the issue comes with the tests. With the changes you made, it looks like some formats may have changed which caused the tests to fail. If you want to touch up the tests with this change we can take a look at how much output would change

xtreme-shane-lattanzio:
FYI I was playing around with thor today and there has been some changes that break LF. I created an issue here: https://github.com/erikhuda/thor/issues/751. Not sure of the plan to fix this but I think this type of thing is why the version of thor was more restrictive :\

nertzy:
It doesn't seem like thor is going to address this problem any time soon.

Would it make more sense to stop relying on thor's `say` and instead directly write to stdout?

clemens:
Is there any intention of fixing this in LicenseFinder since Thor apparently won't change the behavior on their side? From what I can see, depending on Thor < 1.1.0 has the downside of effectively not supporting Ruby 3 (at least the Thor changelog says that 1.1.0 introduced Ruby 3 compatibility and [the change to use `URI.open`](https://github.com/rails/thor/commit/4ce38c5478a5a13f2214ca833eafc167615b4e6a) indicates this as well). While definitely not everyone uses Ruby 3 already, it's the recommended Ruby version for both Rails 6.1 and 7.0.

nertzy:
I think it would be a great idea to use `puts` or something else simple to print to the console instead of `say`. If I get a free moment, I want to look into this.

I now have a project where I can not upgrade to the latest version of [karafka](https://rubygems.org/gems/karafka) because of the thor version restriction.

nertzy:
What can I do to get this merged?

xtreme-shane-lattanzio:
Hey @nertzy ! Sorry for the delay on this. I think you are right saying that this is how thor is now and we should probably do something about it. The change will be quite large but I imagine to be trivial to get `say` out of the code. If we can get the same output for license finder runs but using stdout or puts or something else, I am all for it and we can get the thor version relaxed. If you have time for this, I am willing to work through this with you because I do not have time to do this right now on my own. Let me know if you need help or questions and Ill let you know if I have any concerns. I really do appreciate you looking into this!

xtreme-shane-lattanzio:
I made a workaround to get this out so we didn't make major changes. https://github.com/pivotal/LicenseFinder/pull/886 will update so the newest thor is used. Gonna close this for now!

PR #820, opened 2021-04-07
Set default value of version in package class to empty string

ohlookadollar:
LicenseFinder produces the following error when there are 2 packages with the same name and the version is nil.

```
       expected no Exception, got #<ArgumentError: comparison of LicenseFinder::MergedPackage with LicenseFinder::MergedPackage failed> with backtrace:
         # ./lib/license_finder/report.rb:19:in `sort'
         # ./lib/license_finder/report.rb:19:in `sorted_dependencies'
         # ./lib/license_finder/reports/csv_report.rb:19:in `block in to_s'
         # ./lib/license_finder/reports/csv_report.rb:18:in `to_s'
```

To fix this we set the default value to be empty string.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -27,9 +27,9 @@ BlockLength:
 ClassLength:
   Enabled: false
 CyclomaticComplexity:
-  Max: 10
+  Max: 15
 PerceivedComplexity:
-  Max: 10
+  Max: 15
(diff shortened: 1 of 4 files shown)


julia-pu:
LGTM 👍

xtreme-debbie-chen (approved):

PR #823, opened 2021-04-14
[Changed] exit when go mod list command fails

Pivotal-Jeff-Jun:


lib/license_finder/package_managers/go_modules.rb
```diff
@@ -58,6 +58,7 @@ def packages_info
         go_list_cmd = "GO111MODULE=on go list -mod=readonly -deps -f '#{format_str}' ./..."
         info_output, stderr, status = Cmd.run(go_list_cmd)
```
ohlookadollar, line 59: Would it be beneficial to also show the info_output (stdout) when it fails?

Pivotal-Jeff-Jun: Because the go command errors, `info_output` actually has no value.

ohlookadollar (approved):

Issue #809, opened 2021-01-12
can we get top level dependency only


prathapar:
Instead of getting dependency of dependencies . is there a way we can get only top level dependency only.

Deklin:
Ty @prathapar

Deklin:
Or perhaps a column if this was a direct dependency or a dependency of a dependency

xtreme-shane-lattanzio:
Hey @prathapar and @Deklin ! There is the project_roots function that will show you the folder that the package managers were found in without digging into the files themselves. Other than that I am wondering about the level of depth you actually care about. For example, if you have a Gemfile with 5 gems in it, I assume you are expecting 5 (or 6 including bundler) entries when running license finder? Maybe an example to clarify as each package manager would handle this differently

Deklin:
@xtreme-shane-lattanzio  
an example for node projects

I only want to get the licenses from dependencies directly defined in package.json, I don't want licenses for dependencies of those, so this hypothetical  new flag would prevent the dependencies of dependencies.

topikachu:
@xtreme-shane-lattanzio
It's better if all package manager executables can accept extra command arguments similar to the @maven_options
For example, define an @npm_options so the user is able to send --depth 0 which only gets the first level of npm libs.

My 2 cents.

xtreme-shane-lattanzio:
Okay so if a package.json has 5 lines, you want the 5 deps. I think a depth flag can definitely be useful here. I am also wondering if there is a way to do this with just some kind of npm flag to not pull down too much. That would fix this too so that we only need to pass a new flag to npm

Deklin:
Agree with @topikachu

PR #827, opened 2021-05-11
Added --npm_options option to customize npm behavior

alexmalott:
Just copied the implementation of `maven_options` to allow customization of npm behavior.  In my team's specific use case, we want to pass `--depth=0` to `npm list` so we see only "direct" dependencies.

Addresses #809 (for npm, anyway)


xtreme-shane-lattanzio:
@alexmalott Thanks for the PR. LGTM!

PR #830, opened 2021-05-20
[Added] Zlib License

blooper05:
## Zlib License

https://opensource.org/licenses/Zlib


xtreme-shane-lattanzio:
Hey @blooper05 ! It looks like this has caused some test failures due to [this](https://github.com/pivotal/LicenseFinder/blob/0f004b528d436b4d53db8bd373ede0594c07d9e8/lib/license_finder/license/definitions.rb#L357). Take a look and then we can merge this in!

blooper05:
Hi, @xtreme-shane-lattanzio, thanks for letting me know!
I have now changed the expectation of that test. Please check it 😆

Issue #714, opened 2020-04-29
Pip 20.1 internal API change


dmcauslan:
Good morning,

I caught this error today: 
```
pip3 install: LicenseFinder command 'python3 /var/lib/gems/2.5.0/gems/license_finder-6.1.2/bin/license_finder_pip.py /app/requirements.txt' failed:
	Traceback (most recent call last):
  File "/var/lib/gems/2.5.0/gems/license_finder-6.1.2/bin/license_finder_pip.py", line 26, in <module>
    if req.req == None or (req.markers != None and not req.markers.evaluate()): continue
AttributeError: 'ParsedRequirement' object has no attribute 'req'
```

It looks like pip 20.1 changed the internal api for `parse_requirements` to return a list of `ParsedRequirement` rather than `InstallRequirement`.  The new object has a `requirement` field while the old object had a `req` field.  The short term fix is pinning pip to < 20.1, but this will need a fix:

https://github.com/pivotal/LicenseFinder/blob/2f79904a8e3fcd02922638dbcb3258960523d7d9/bin/license_finder_pip.py#L25

agconti:
Here are steps I needed to work around this issue:

Create and activate a virtualenv. Downgrade the version of pip within that env:
```
pip install pip==20.0.2  
```

Install dependencies:
```
pip3 install -r requirements.txt
```

Run license finder:
```
license_finder report --python-version=3
```

olliemath:
Hello, any progress on this? Pip 20.0.2 is quite old now.

It should be noted that the above work-around won't work for any installs that require the newer features of pip. I'm happy to open a PR with the required change if you need somebody to fix the offending line in `license_finder_pip.py`

xtreme-shane-lattanzio:
Hi @olliemath ! We would appreciate a PR! We have no spent the time to make the backwards compatibility which is why we recommend the downgrade for now but if you can make the change, we would be more than happy to take a look and merge it in. Thanks!!!

olliemath:
OK, see attached PR

olliemath:
Hello, did anyone have a chance to look at my PR? It's been open for a month.

PR #833, opened 2021-06-03
Compatibility: license_finder_python for pip 20.1+

olliemath:
The API for pip changed with version 20.1. This commit introduces the
necessary compatibility hacks.

I couldn't see any obvious python tests in your ci setup. Manually tested with versions 19.x, 20.x, 21.x and the following requirements file under python 3.7:
```python
flask==1.0.0
six==1.12.0 ; python_version < 3.4
mock==1.3.0 ; python_version < 3.8
```

Fixes https://github.com/pivotal/LicenseFinder/issues/714


xtreme-shane-lattanzio:
@olliemath Thanks for this! We have known about this for a while and haven't fixed it so this is a big help!

Issue #838, opened 2021-06-30
Running license_finder with bundle install --path results in gems not being found


Fire-Dragon-DoL:
Greetings.
After installing license_finder with `gem install license_finder` and running `bundle install --path vendor/bundle` in my project, `license_finder` seems to fail (can't find gems).

I digged a bit in the issue tracker and it seems to be a problem that has been fixed, but even with the latest version (6.14.1) I'm still facing the same issue.

Notice that if no path is specified, the problem does not happen.

Any guidance on how I may debug this issue and resolve it?

Fire-Dragon-DoL:
Notice that I think the issue might be the same mentioned here: https://github.com/pivotal/LicenseFinder/issues/828

```
Bundler version 2.2.21 (2021-06-23 commit 0b5db72094)
ruby 2.7.3p183 (2021-04-05 revision 6847ee089d) [x86_64-linux]
license_finder version 6.14.1
```

Fire-Dragon-DoL:
Another note, I wonder if  the problem presents when the gem (license_finder) has a dependency with a version colliding with the one in the gem file.

For example, my Gemfile.lock requires thor 0.20.1 but license_finder uses thor 1.x.x.

I'll perform further tests.

Fire-Dragon-DoL:
I think there might be a discussion to be made about the design for bundler.

The idea being since this is a ruby process, it should load Bundler natively (call it in Ruby with `Bundler.xxx`) to determine the libraries and their licenses.

I'd argue that this is the root cause of the problem and bundler should be treated like every other package manager: a CLI tool which should be interacted with only over CLI.

I'm writing an experimental branch with a new package manager (BundlerCLI) that does exactly that. Based on my exploratory testing, this fixes the issue because it runs bundler in a "subshell" essentially, so it doesn't affect LF execution.

optimistic5:
Any updates on it?
It would be great if license_finder will understand `bundle config set path 'vendor'` and will check dependencies there.

Issue #828, opened 2021-05-12
Could not find addressable-2.4.0 in any of the sources


chatziparaskewas:
While license_finder is working nicely on Maven, NPM, etc. projects. Something triggers the below problem one way or the other. This happens with LF running from the command line (gem install), via Gemfile+bundle exec or via Docker.

Unfortunately, I cannot attach the project causing the issue.

The log below is from running Gemfile+bundle exec. The file /usr/bin/license_finder is physically existing.
Trying to add addressable-2.4.0 manually (gem install) does not resolve the issue. This is only the first missing dependency of many more and finally it ends in a dead end (I tried Ruby 2.5.5 and 3.0.1).

license_finder: 6.13.0
Ruby: 2.5.5
OS: CentOS Linux release 8.3.2011

------------------------------------
LicenseFinder::GoModules: is not active
LicenseFinder::GoDep: is not active
LicenseFinder::GoWorkspace: is not active
LicenseFinder::Go15VendorExperiment: is not active
LicenseFinder::Glide: is not active
LicenseFinder::Gvt: is not active
LicenseFinder::Govendor: is not active
LicenseFinder::Trash: is not active
LicenseFinder::Dep: is not active
LicenseFinder::Bundler: is active
LicenseFinder::NPM: is not active
LicenseFinder::Pip: is not active
LicenseFinder::Yarn: is not active
LicenseFinder::Bower: is not active
LicenseFinder::Maven: is not active
LicenseFinder::Gradle: is not active
LicenseFinder::CocoaPods: is not active
LicenseFinder::Rebar: is not active
LicenseFinder::Erlangmk: is not active
LicenseFinder::Nuget: is not active
LicenseFinder::Carthage: is not active
LicenseFinder::Mix: is not active
LicenseFinder::Conan: is not active
LicenseFinder::Sbt: is not active
LicenseFinder::Cargo: is not active
LicenseFinder::Dotnet: is not active
LicenseFinder::Composer: is not active
LicenseFinder::Pipenv: is not active
LicenseFinder::Conda: is not active
LicenseFinder::Spm: is not active
LicenseFinder::Bundler: is installed
LicenseFinder::Bundler: including groups [:test]
bundler: failed to load command: license_finder (/usr/bin/license_finder)
Traceback (most recent call last):
	40: from /home/xxx/bin/bundle:23:in `<main>'
	39: from /home/xxx/bin/bundle:23:in `load'
[...]

PR #839, opened 2021-06-30
Bundler CLI package manager is now present

Fire-Dragon-DoL:
This PR is to initiate a discussion.
It adds a new Package manager called "Bundler CLI" which as suggested, uses bundler as a CLI tool, not as a ruby library.
This approach treats bundler like every other package manager, eliminating a big category of errors caused by loading bundler in the same process where `license_finder` is executed. Notice this problem can't happen for non-ruby package managers.

An alternative approach to this would be forking within license finder and only then execute Bundler-related operations, with some IPC mechanism to return the result.

This CLI approach has the downside of paying the "ruby initiation" cost repeatedly, which results in some serious slowdowns when dealing with a lot of dependencies. Using `Parallel` provides substantial relief, at the cost of spinning fans.

I wrote this code after encountering a blocking bug (https://github.com/pivotal/LicenseFinder/issues/838) when dealing with `bundle install --path something`, where `license_finder` would attempt to load bundler and fail to find any gems because not executed with bundle exec, but it would also fail with bundle exec because it's not part of the bundle and cannot be part of the bundle due to a dependency conflict.

## TODO

- [ ] Help update
- [ ] Tests
- [ ] Support `--bundler-options`


xtreme-shane-lattanzio:
Hey @Fire-Dragon-DoL ! Thanks for this work. The bundler pathing issues is something that has always plagued us. It has gone through many different iterations to try to solve all cases. We have environment variables for gemfile path and have tried to customize the gem_path in the prepare command. I think https://github.com/pivotal/LicenseFinder/issues/828 may be related and he found a workaround for the gem path.

Normally I would be against this alternative kind of approach but since there are so many issues with this, I would be fine with adding this as some kind of `beta` feature. Have you looked into the gem_path stuff on line 30 of bundler.rb? Or is that not close to this problem?

Fire-Dragon-DoL:
hey @xtreme-andy-shen thanks for reading my PR.
I tried really hard to get that approach working, but I was already running (manually) `bundle install --vendor/bundle`, which made that fix irrelevant (although I did try it).

I think that in most cases, the problem is that those tests are performed in a state where the gems are also installed globally first (raw `bundle install`) and then `bundle install --path`, so the gems are in both locations.
I can't confirm this though.

That being said, bundler is built in a way intended to alter the _current_ (global) ruby environment so that it's isolated from the rest of the world, given this premise, it makes sense that LF is having so many problems with it.

The approach I used is "ugly" for a ruby package (should be able to invoke it from within Ruby), but seen from the perspective of the other package managers, it makes sense: they are all used through CLI.

If we want to use bundler directly and avoid issues, I think the ideal approach is to isolate bundler within Ruby: that is, fork.
This will separate and isolate the two processes, so it should be possible to run bundler in one without "breaking" LF.

This PR is indeed intended as beta. It works "flawlessly" (I'm the only one who tested it, so I can say very little about it) with bundler though, so it can solve the problem for a lot of people while we get to a better state.
The core problem is identifying the better state: is **fork** the right approach? If not, what would it be?
I was hoping to discuss this so that I can improve the PR.

Cheers!

Fire-Dragon-DoL:
hey @xtreme-shane-lattanzio thanks for reading my PR.
I tried really hard to get that approach working, but I was already running (manually) `bundle install --vendor/bundle`, which made that fix irrelevant (although I did try it).

I think that in most cases, the problem is that those tests are performed in a state where the gems are also installed globally first (raw `bundle install`) and then `bundle install --path`, so the gems are in both locations.
I can't confirm this though.

That being said, bundler is built in a way intended to alter the _current_ (global) ruby environment so that it's isolated from the rest of the world, given this premise, it makes sense that LF is having so many problems with it.

The approach I used is "ugly" for a ruby package (should be able to invoke it from within Ruby), but seen from the perspective of the other package managers, it makes sense: they are all used through CLI.

If we want to use bundler directly and avoid issues, I think the ideal approach is to isolate bundler within Ruby: that is, fork.
This will separate and isolate the two processes, so it should be possible to run bundler in one without "breaking" LF.

This PR is indeed intended as beta. It works "flawlessly" (I'm the only one who tested it, so I can say very little about it) with bundler though, so it can solve the problem for a lot of people while we get to a better state.
The core problem is identifying the better state: is **fork** the right approach? If not, what would it be?
I was hoping to discuss this so that I can improve the PR.

Cheers!

Fire-Dragon-DoL:
@xtreme-shane-lattanzio This is an alternative approach to what I provided here. It's substantially faster and accesses Bundler through ruby, however it does so by running a separate ruby executables. Overall, the downsides seem minimal and solves an existing bug: https://github.com/pivotal/LicenseFinder/pull/843

xtreme-shane-lattanzio:
Closing in favour of https://github.com/pivotal/LicenseFinder/pull/843

PR #853, opened 2021-08-28
Removed rewriting for local go modules

ohlookadollar:


timhaines:
@ohlookadollar I hadn't yet learnt replacement modules existed in Go. Interesting feature.  Why are you thinking of removing them from a license analysis perspective?

ohlookadollar:
@timhaines we noticed for projects rewrite modules with local modules. Ex. `replace github.com/example/test ../../test` the dependency LF returns '../../test' which isn't helpful. The change does ensures the LF returns github.com/example/test. 

This change also affects non local module replacements as well. Ex. `replace github.com/NetEase-Object-Storage/nos-golang-sdk => github.com/karuppiah7890/nos-golang-sdk v0.0.0-20191116042345-0792ba35abcc`.

ohlookadollar:
We might not want this to be merged in as it affects non local packages from being rewritten

PR #855, opened 2021-08-30
Improve CocoaPods support: multiple targets, licenses from specs, subspec inheritance

reidab:
This addresses two issues with the existing implementation:

1. License information was only loaded from the first acknowledgements path. In projects with multiple build targets, this resulted in a failure to identify many dependencies.
2. License names provided in the acknowledgements were ignored and the license text was always used to determine a license. This refactors which data is passed to `CocoaPodPackage` to allow the license names to be read from the spec, falling back to searching by text.

It also adds fallback behavior for subspec entries, allowing them to look up their license from their parent package.

Unlike other package managers like npm where a slash in a package name denotes a namespace across multiple package like `@babel/something` in npm, CocoaPods subspecs are specified in a single podspec file.
(e.g. https://github.com/CocoaPods/Specs/blob/master/Specs/3/2/5/FlipperKit/0.104.0/FlipperKit.podspec.json)

Per https://guides.cocoapods.org/syntax/podspec.html#subspec, the intended behavior is that "a ‘sub-specification’ inherits the value of the attributes of the parents so common values for attributes can be specified in the ancestors."

I believe this means that we can safely fall back to the parent specification to look up a license value for subspecs when the subspec does not provide one of its own.

### DIFF
--- a/lib/license_finder/package_managers/cocoa_pods.rb
+++ b/lib/license_finder/package_managers/cocoa_pods.rb
@@ -15,7 +15,7 @@ module LicenseFinder
         CocoaPodsPackage.new(
           name,
           version,
-          license_texts[name],
+          acknowledgements[name],
           logger: logger
         )
       end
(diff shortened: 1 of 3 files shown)


xtreme-shane-lattanzio:
Hey @reidab . Thanks for this PR! The cocoapods section has definitely been lacking some love. It looks like your changes broke some tests so if you can touch those up, I can look into merging this, Thanks again!!

xtreme-shane-lattanzio:
Hey @reidab ! Just wanted to bump this as I do think it is valuable to have merged in. Would you be able to touch up the tests and rebase so we can get this merged in? Let me know when you have a chance!

xtreme-shane-lattanzio:
@reidab I just merged https://github.com/pivotal/LicenseFinder/pull/873/files which seems to be similar to what you were doing here. Can you take a look and let me know if it helps? Thanks!

xtreme-shane-lattanzio:
Closing in favour of https://github.com/pivotal/LicenseFinder/pull/873/

PR #857, opened 2021-09-11
Safely load dependency decisions with any version of Psych

### LINKED ISSUE #841: Error when running with Psych 4.0
When running `license_finder` in a project that has some dependency decisions, with Psych 4.0 installed, the following error happens:

```
[...]/lib/ruby/gems/3.0.0/gems/psych-4.0.1/lib/psych/class_loader.rb:99:in `find':
  Tried to load unspecified class: Time
```

mvz:
Psych 4 no longer loads YAML unsafely with `Psych.load`. This change switches to `Psych.safe_load`. Since the calling convention for `Psych.safe_load` was changed, and the old convention is no longer supported in Psych 4, either the old or the new convention is used based on the loaded version of Psych.

Fixes #841.

### DIFF
--- a/lib/license_finder/decisions.rb
+++ b/lib/license_finder/decisions.rb
@@ -280,7 +280,11 @@ module LicenseFinder
     def self.restore(persisted, result = new)
       return result unless persisted
 
-      actions = YAML.load(persisted)
+      actions = if Psych::VERSION < '3.1.0'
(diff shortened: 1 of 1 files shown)


xtreme-jason-smith:
Duplicate #877

mvz:
Not sure what to think here ...

PR #864, opened 2021-12-02
[Changed] Sanitize summary and description in HTML report

surc54:
When creating an HTML license report, I noticed that the description of the [markupsafe](https://github.com/pallets/markupsafe) library caused Javascript to execute on the browser.

To fix this, I changed the HTML report to sanitize the summary and description.


xtreme-shane-lattanzio:
Hey @surc54 ! Thanks for putting this in. Unfortunately as per https://github.com/rgrove/sanitize/releases, santize version 6.0 only supports ruby 2.5 and up and we currently support down to ruby 2.3. If you want to change this to version [5.2.3](https://github.com/rgrove/sanitize/releases/tag/v5.2.3), we can merge it in. Otherwise we may need to hold off on this

xtreme-shane-lattanzio:
Closing due to inactivity

PR #873, opened 2022-01-19
:wrench: allow to specify cocoadpods acknowledgment through ENV

etiennecadicidean:
> in multi projects folder of iOS projects, it looks like the cocoapods plugin is unable
to find the proper Pods-...-acknowledgments.plist file. As it takes only the 1st item of the dir,
We should be able to also specify the one we want trought an environment variable ACKNOWLEDGEMENT_PATH
just like what it is done in the SPM plugin where we do need to specify the environment variables for
DerivedDatas & Package.resolved files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ languages, as long as that language has a package definition in the project dire
 * `build.gradle` (for `gradle`)
 * `settings.gradle` that specifies `rootProject.buildFileName` (for `gradle`)
 * `bower.json` (for `bower`)
-* `Podfile` (for `pod`)
+* `Podfile` (for `pod`) (set `ACKNOWLEDGEMENTS_PATH` variable if you want to target a particular `Pods-acknowledgements-<TARGET>.plist`. Can be useful in multi-target pods projects.)
 * `Cartfile` (for `carthage`)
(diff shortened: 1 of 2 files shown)


etiennecadicidean:
Not sure why the bundle install step fails for some jruby tests on 

concourse-ci/ruby-jruby-9.2.14.0 

it has nothing to do with my minor changes and I guess it fails on the main branch as well.

xtreme-jason-smith:
Hi @etiennecadicidean , the PR looks good overall.  The only change I would suggest is expose this new environment variable, either though the README, or through a raise error (similar to spm), or both. We want to make sure everyone is aware of this new feature.

Do not worry about jruby as that is not a blocking test as we have not had the bandwidth yet to investigate the failing tests.

etiennecadicidean:
> Hi @etiennecadicidean , the PR looks good overall. The only change I would suggest is expose this new environment variable, either though the README, or through a raise error (similar to spm), or both. We want to make sure everyone is aware of this new feature.

@xtreme-jason-smith as the variable is completely optional and can help in very specific situations I only added additional informations in the README.md ! Thank you for your cutting comments on all my PRs 👍

xtreme-shane-lattanzio:
Hey @etiennecadicidean ! Dont worry about the jruby stuff. It has been failing for a while and we will be removing it from our pipelines soon. We are just fixing up our CI and will merge this in. Thanks again!!

PR #875, opened 2022-01-26
Provide new package manager support : Pub for Flutter projects

etiennecadicidean:
Provide new package manager support : Pub for Flutter projects


Dockerfile
```diff
@@ -206,6 +206,25 @@ RUN apt-get -q install -y \
     pkg-config \
     && rm -r /var/lib/apt/lists/*
 
+#install flutter
+ENV FLUTTER_HOME=/root/flutter
+RUN curl -o flutter_linux_2.8.1-stable.tar.xz https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_2.8.1-stable.tar.xz
+RUN tar xf flutter_linux_2.8.1-stable.tar.xz
+RUN mv flutter ${FLUTTER_HOME}
+RUN rm flutter_linux_2.8.1-stable.tar.xz
```
xtreme-jason-smith, line 214: These RUN commands should be merged into a single RUN to minimize the number of docker layers and docker size. Unless there is a reason these need to be ran as separately?

spec/lib/license_finder/package_managers/pub_package_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+        end
+      end
+
+      #   context "when spm did not run yet" do
+      #     it "raises an exception to explain the reason" do
+      #       expect do
+      #         spm.current_packages.first.licenses.map(&:name)
+      #       end.to raise_exception(Spm::SpmError)
+      #     end
+      #   end
```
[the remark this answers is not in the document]
etiennecadicidean: no good reason for it : was used as inspirational from spm tests as at 1st look, the pms were looking similar. Now tests are written, no good reason to keep it commented 👍 removed it.

Dockerfile
etiennecadicidean: Definitely true. Had to separate to help debugging but now, no good reason to keep it that way. I fixed it. Thank you !

etiennecadicidean:
Hey there, just to let you know I wrote an article on how to monitor licenses with Sonarqube using Licensefinder to generate some of the reports (iOS/Flutter). For now, we link to our own gem (https://github.com/etiennecadicidean/LicenseFinder) because it has all the latest features in the PRs I created, but I would be most happy if I could link to your own repository instead). 

There will be a frog/capgemini communication about the article tomorrow at midday (UTC+1 time), so if the PR could be merged before, I could amend the article to this repository, I think it would benefit to everyone 👍 

The article : 
https://medium.com/@etienne.cadic_60777/opensource-licenses-monitoring-with-sonarqube-4783785a033d

Thank you !

xtreme-shane-lattanzio:
@etiennecadicidean Was just waiting for CI and now merged! Thanks again!!

PR #876, opened 2022-02-01
:sparkle: add a new option to use spdx identifier in reports

etiennecadicidean:
- SPDX identifier is essential to make license match between CI tools.
 - Using the proper --use_spdx_licenses, the user can get those licenses
in the proper format to use reports in tools such as Sonar License Check plugins

### DIFF
--- a/lib/license_finder/license/definitions.rb
+++ b/lib/license_finder/license/definitions.rb
@@ -46,7 +46,7 @@ module LicenseFinder
         License.new(
           short_name: 'Apache1_1',
           pretty_name: 'Apache 1.1',
-          spdx_id: 'Apache-1.0',
+          spdx_id: 'Apache-1.1',
           other_names: [
             'Apache-1.1',
             'The Apache Software License, Version 1.1'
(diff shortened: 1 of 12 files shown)


etiennecadicidean:
Hello there @xtreme-shane-lattanzio 
the CI fails tests while everything is fine in the docker image using : 

```
./dlf rake spec
./dlf bundle exec rake features

```
Can you figure out what it takes to make the PR approved ?

lib/license_finder/license/definitions.rb
```diff
@@ -46,6 +46,7 @@ def apache1_1
         License.new(
           short_name: 'Apache1_1',
           pretty_name: 'Apache 1.1',
+          spdx_id: 'Apache-1.0',
```
xtreme-debbie-chen, line 49: Is this correct? This is the license for Apache 1.1. I would expect the spdx id to be  `Apache-1.1` (from the [spdx license list](https://spdx.org/licenses/))

etiennecadicidean: totally right, I just fixed it. Thank you for reviewing !

etiennecadicidean: i doubled check the other ones, should be okay.

xtreme-shane-lattanzio:
@etiennecadicidean Was just waiting for CI and now merged! Thanks for another one!!

Issue #872, opened 2022-01-18
Yarn2 fails silently and does not find any dependencies/licenses in a monorepo without workspaces defined


nickbreen:
Command line: `report --prepare --recursive --enabled-package-managers yarn`

In our monorepo we have a number of independent yarn projects each with their own `package.json` and `yarn.lock` file, _we have no root `package.json`/`yarn.lock` and we're not using workspaces_. 

License finder logs:

> LicenseFinder::Yarn: is active for /foo
> LicenseFinder::Yarn: is active for /bar
> LicenseFinder::Yarn: is active for /baz

But finds no dependencies or licenses!

Inspection of [yarn.rb](https://github.com/pivotal/LicenseFinder/blob/cfbc7d3fd5313b4f19df523293b01280aa52ef65/lib/license_finder/package_managers/yarn.rb) shows the prepare command executed is:

     yarn plugin import workspace-tools && yarn workspaces focus --all --production && yarn install

When run manually that command fails on the first command `yarn plugin import` with:

> YN0001: UsageError: No project found in /buck-all

Due to there being no root-project.

**Problem 1** is that this error is _silent_. Even with `--debug` enabled there is no indication that the command has failed, the report is 'successfully' generated with zero dependencies and the tool exits with a zero exit code indicating success.

**Problem 2** is that license finder is capable of _finding_ the yarn projects but then fails to execute against the found project.

nickbreen:
A work-around is to use `project-roots --recursive` and use the output to generate the command line for `report --aggregate-paths ...` which treats each path as a discrete project.

nickbreen:
Also even if the 'production' prefix for the `yarn install` command is empty the `yarn install` command will fail also due to there being no root-project/workspace.

xtreme-shane-lattanzio:
Hey @nickbreen ! Thanks for raising this. There seems to be two issues here. One is the prepare command not working when there is no root project. And the other is about silent failing when you would prefer it to actually error out. Do you have proposed solutions for this? The failing out is easy enough but Im wondering if you paths work around for the other issue is good enough? We welcome any PRs if you have an idea for this :)

PR #879, opened 2022-02-15
[Changed] Raise an error if the yarn licenses list command fails

ajesler:
Previously, LicenseFinder::Yarn.new.current_packages silently returned an empty array when `yarn licenses list` failed, resulting in no license violations being detected.
This is dangerous as it can lead to thinking there are no license issues present.

Missing or incorrectly configured environment variables can lead to
`yarn licenses list` failing.

The change does break backwards compatibility, but I'm struggling to think of a case where you'd want silent failures.


xtreme-shane-lattanzio:
Hey @ajesler ! Thanks for the PR. This should partially fix https://github.com/pivotal/LicenseFinder/issues/872#issuecomment-1047929085. However,  it looks like your changes broke rubocop formatting. Please run rubocop on your changes and then we can hopefully get past tests and merge this in!

PR #885, opened 2022-03-01
Remove pinning for thor

dbussink:
This is pinning to a really old release of `thor` that results in a downgrade when upgrading `license_finder` for us.


dbussink:
Whoops, missed #880

PR #890, opened 2022-03-09
[Fixed] Use relative path to working directory for maven wrapper command

### LINKED ISSUE #889: Fail to detect Maven when using Maven Wrapper with aggregate paths on a Maven project in a subdirectory
We are currently evaluating License Finder at work and we are facing an issue with Maven project in a subdirectory when scanning multiple projects using the `--aggregate_paths` option.

To simply and quickly evaluate License Finder, we are using the Docker image in interactive mode (`docker run -v $PWD:/scan -it licensefinder/license_finder`).

In our case we have a mono repository that is organized this way (_simplified/redacted view_):
```
project/
├── backend
│   ├── mvnw
│   ├── mvnw.cmd
│   └── pom.xml
├── frontend
│   └── package.json
└── package.json
```

The `project` directory is mounted at the path `/scan` and we run License Finder with `/LicenseFinder/bin/license_finder report --aggregate_paths='.' './frontend/' './backend/'`

Scanning NPM projects located at `.` and `./frontend/` works fine.

But scanning the Maven project (with submodules) located at `./backend/` fails with the following:
```
LicenseFinder::Maven: is active
LicenseFinder::Maven: is not installed
```

This is because it is not able to find the `./mvnw` command from the root directory of the project. Indeed that file is located in the `backend` directory.

In `PackageManager` it checks that a [command exists](https://github.com/pivotal/LicenseFinder/blob/791faf11634ac71451b8e32959a314cd1e73e08e/lib/license_finder/package_manager.rb#L54):
```ruby
    def command_exists?(command)
      _stdout, _stderr, status =
        if LicenseFinder::Platform.windows?
          Cmd.run("where #{command}")
        else
          Cmd.run("which #{command}")
        end

      status.success?
    end
```

And in `Maven` the command is [defined](https://github.com/pivotal/LicenseFinder/blob/e2fd2895f919585da13b02456b77ecced75c3a98/lib/license_finder/package_managers/maven.rb#L36) as:
```ruby
    def package_management_command
      wrapper = if Platform.windows?
                  'mvnw.cmd'
                else
                  './mvnw'
                end
      maven = 'mvn'

      File.exist?(File.join(project_path, wrapper)) ? wrapper : maven
    end
```

In the case of Maven Wrapper the path to the command should not be relative (`./mvnw`) but absolute w.r.t `project_path`.

We tested with the following and it works as expected:
```ruby
    def package_management_command
      wrapper = File.join(project_path, Platform.windows? ? 'mvnw.cmd' : 'mvnw');
      maven = 'mvn'

      File.exist?(wrapper) ? wrapper : maven
    end
```

We will shortly create a PR to fix that issue.

jbmgrtn:
Resolves #889

### DIFF
--- a/lib/license_finder/package_managers/maven.rb
+++ b/lib/license_finder/package_managers/maven.rb
@@ -34,14 +34,10 @@ module LicenseFinder
     end
 
     def package_management_command
-      wrapper = if Platform.windows?
-                  'mvnw.cmd'
-                else
-                  './mvnw'
-                end
+      wrapper = File.join(project_path, Platform.windows? ? 'mvnw.cmd' : 'mvnw');
       maven = 'mvn'
 
-      File.exist?(File.join(project_path, wrapper)) ? wrapper : maven
+      File.exist?(wrapper) ? wrapper : maven
     end
 
     def possible_package_paths
--- a/spec/lib/license_finder/package_managers/maven_spec.rb
+++ b/spec/lib/license_finder/package_managers/maven_spec.rb
@@ -34,9 +34,9 @@ module LicenseFinder
 
       it 'uses the maven wrapper, if present' do
         subject = Maven.new(project_path: Pathname('features/fixtures/maven-wrapper'))
-        allow(SharedHelpers::Cmd).to receive(:run).with('./mvnw org.codehaus.mojo:license-maven-plugin:download-licenses').and_return(['', '', cmd_success])
+        allow(SharedHelpers::Cmd).to receive(:run).with('features/fixtures/maven-wrapper/mvnw org.codehaus.mojo:license-maven-plugin:download-licenses').and_return(['', '', cmd_success])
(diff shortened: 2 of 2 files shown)


xtreme-shane-lattanzio:
Hey @jbmgrtn and thanks for the PR! This looks good to me. As for ruby code cleanup, we actually run rubocop to do that. You can run `rubocop` in the root directory to see whats wrong or `rubocop -a` to have the issues auto fixed. Looking at the pipeline, the only issue is 

```Inspecting 278 files
Offenses:

lib/license_finder/package_managers/maven.rb:37:81: C: [Correctable] Style/Semicolon: Do not use semicolons to terminate expressions.
      wrapper = File.join(project_path, Platform.windows? ? 'mvnw.cmd' : 'mvnw');
```

I think the fix is trivial enough but to be safe, is it an issue to sign the CLA? It is basically just saying you do not own the change and are donating it to the project.

Once we have the pipeline green I'll merge this in!

PR #891, opened 2022-03-10
Support yarn licenses command in yarn v2+

julia-pu:


lib/license_finder/package_managers/yarn.rb
```diff
@@ -2,16 +2,20 @@
 
 module LicenseFinder
   class Yarn < PackageManager
-    SHELL_COMMAND = 'yarn licenses list --no-progress --json'
```
Pivotal-Jeff-Jun, line 5: is `--no-progress` a deprecated flag? or we don't need it anymore?

lib/license_finder/package_managers/yarn.rb
```diff
@@ -2,16 +2,20 @@
...
 
     def current_packages
+      #keeping the use of yarn1_production_flag here because the license plugin supports same commands as licenses in yarn v1
       cmd = "#{Yarn::SHELL_COMMAND}#{yarn1_production_flag}"
-      suffix = " --cwd #{project_path}" unless project_path.nil?
-      cmd += suffix unless suffix.nil?
+      unless yarn2_project?
+        cmd += " --no-progress"
+        suffix = " --cwd #{project_path}" unless project_path.nil?
+        cmd += suffix unless suffix.nil?
```
Pivotal-Jeff-Jun, line 17: could we refactor these two lines into one?
```ruby
cmd += " --cwd #{project_path}" unless project_path.nil?
```

lib/license_finder/package_managers/yarn.rb
Pivotal-Jeff-Jun: or is that this flag is only in yarn 2 and not 1

lib/license_finder/package_managers/yarn.rb
```diff
@@ -83,6 +92,16 @@ def yarn2_project?
       end
     end
 
+    def yarn3_project?
```
xtreme-shane-lattanzio, line 95: This should be refactored to just return the project type. Duplicating the same code and then checking for a number isnt great

lib/license_finder/package_managers/yarn.rb
```diff
@@ -2,16 +2,20 @@
...
   class Yarn < PackageManager
-    SHELL_COMMAND = 'yarn licenses list --no-progress --json'
+    SHELL_COMMAND = 'yarn licenses list --json'
 
     def possible_package_paths
       [project_path.join('yarn.lock')]
     end
 
     def current_packages
+      #keeping the use of yarn1_production_flag here because the license plugin supports same commands as licenses in yarn v1
```
xtreme-shane-lattanzio, line 12: Should this flag be renamed then?

lib/license_finder/package_managers/yarn.rb
```diff
@@ -66,7 +70,12 @@ def prepare_command
     private
 
     def yarn2_prepare_command
```
xtreme-shane-lattanzio, line 72: The prepare_command on line 62 only checks for yarn1 or 2. What is it supposed to be if it is yarn 3? I also don't like how this yarn2 prepare command is for yarn 3 as well. This file needs to be refactored

xtreme-shane-lattanzio (changes requested):
This should be cleaned up. Let me know if you wanna chat

lib/license_finder/package_managers/yarn.rb
julia-pu: that flag is only supported in yarn 1, so I'm removing it from the base shell command and appending it only if it's yarn v1

lib/license_finder/package_managers/yarn.rb
julia-pu: the check for yarn2 actually is checking for any versions greater or equal to yarn2, so it includes yarn 3. I was originally trying to change as little things as possible, but I agree it is confusing. I'll do some refactoring

julia-pu: made the refactor

lib/license_finder/package_managers/yarn.rb
julia-pu: changed in the refactor

lib/license_finder/package_managers/yarn.rb
julia-pu: refactored

lib/license_finder/package_managers/yarn.rb
julia-pu: refactored

julia-pu:
Refactored, please take a look again!

xtreme-shane-lattanzio (approved):
Much better. Thank you!!!

xtreme-debbie-chen (approved):

PR #893, opened 2022-03-19
Added the `who` and `why` of manually approved licenses to the CSV report

jonathanbp:
The `who` is represented by the `approved_by` column and the `why` by the `approved_reason`.


xtreme-shane-lattanzio:
@jonathanbp Thanks for the PR! Seeing how simple the feature is confuses me why we didn't do it in the first place. Looks good!

PR #905, opened 2022-05-09
Add NPM authors and contributors as authors

rhuitl:
Reads the author and the list of contributors and writes their name into the "authors" output field.


xtreme-shane-lattanzio:
@rhuitl Sorry for the delay and thanks for your PR. Looks good!

PR #907, opened 2022-05-11
Maven: extract LICENSE and NOTICE texts from JAR + add summary, description, url from POM

### LINKED ISSUE #822: License text extraction not working for Maven
I would like to extract the license texts for Maven projects. What works for NPM does not for Maven. The license text field remains empty.

Steps to reproduce:
1. Create a basic Maven project
```
mvn archetype:generate -DgroupId=com.mycompany.app -DartifactId=my-app -DarchetypeArtifactId=maven-archetype-quickstart -DarchetypeVersion=1.4 -DinteractiveMode=false
```
2. Run LicenseFinder
```
docker run -v "$PWD/my-app:/scan" -t licensefinder/license_finder /bin/bash -l -c "cd /scan && license_finder report --quiet --format json --columns name version authors licenses license_links approved summary description homepage install_path package_manager texts notice"
```

The output is:
```json
{
	"dependencies": [{
		"name": "hamcrest-core",
		"version": "1.3",
		"authors": "",
		"licenses": ["New BSD"],
		"license_links": "http://opensource.org/licenses/BSD-3-Clause",
		"approved": "Not approved",
		"summary": "",
		"description": "",
		"homepage": "",
		"install_path": null,
		"package_manager": "Maven",
		"texts": "",
		"notice": ""
	}, {
		"name": "junit",
		"version": "4.11",
		"authors": "",
		"licenses": ["Common Public License Version 1.0"],
		"license_links": "",
		"approved": "Not approved",
		"summary": "",
		"description": "",
		"homepage": "",
		"install_path": null,
		"package_manager": "Maven",
		"texts": "",
		"notice": ""
	}]
}
```

3. Compare with downloaded licenses
```
❯ find my-app/target/generated-resources/licenses
my-app/target/generated-resources/licenses
my-app/target/generated-resources/licenses/common public license version 1.0 - cpl1.0.html
my-app/target/generated-resources/licenses/new bsd license - bsd-license.html
```

I was trying to understand how the license text extraction works for NPM, and it seems like it's based on the filename (like LICENSE or similar).

I was trying to understand how the Maven module for LicenseFinder works but I'm lacking context and Ruby knowledge to really understand how it's supposed to work - after all it is downloading the license files, so it seems feasible to read in the texts from there.

Happy for any pointers :)

rhuitl:
I'd like to propose this change to the Maven extraction. It adds support for extracting the `texts` and `notice` for Maven. This resolves #822.

Further, we will now extract the following fields from the POM file to the respective LicenseFinder fields: `name` => `summary`, `description` => `description`, `url` => `homepage`.

It extracts the license and notice texts from the JAR files which are downloaded by Maven into `~/.m2`. I found this to be the only way how we can get the license texts (the files downloaded by the Maven plugin are just the license templates, with no author names). I had to make the filename matching for notice files a bit stricter (`notice*`) than your default (`*notice*`) because we do not verify them further as we do for the license files and I got wrong files included (like `.class` files).

If we want to extend the test suite to cover this new functionality, I'd be happy for any hints. Do we actually run Maven somewhere in the test suite already for an end-to-end test? The existing tests seem to stub out Maven, and I'm not sure how well I can cover this code that way.

### DIFF
--- a/features/features/package_managers/maven_spec.rb
+++ b/features/features/package_managers/maven_spec.rb
@@ -21,4 +21,21 @@ describe 'Maven Dependencies' do
       expect(java_developer).to be_seeing_line 'junit:junit, 4.13.1, "Eclipse Public License 1.0"'
     end
   end
+
+  it 'extracts name/description/url from POM and license/notice from JAR' do
+    LicenseFinder::TestingDSL::MavenProject.create
+    java_developer.execute_command 'license_finder report --columns summary description homepage texts notice --format=csv'
+
+    expect(java_developer).to be_seeing_once 'Hamcrest Core,'
+    expect(java_developer).to be_seeing_once 'This is the core API of hamcrest matcher framework'
+    expect(java_developer).to be_seeing_once 'Copyright (c) 2000-2006, www.hamcrest.org'
+
+    expect(java_developer).to be_seeing_once 'JUnit,'
+    expect(java_developer).to be_seeing_once '"JUnit is a unit testing framework for Java'
+    expect(java_developer).to be_seeing_once 'http://junit.org'
+
+    expect(java_developer).to be_seeing_once 'Apache Commons Lang,"Apache Commons Lang, a package of Java utility classes'
+    expect(java_developer).to be_seeing_once 'Version 2.0, January 2004' # LICENSE
+    expect(java_developer).to be_seeing_once 'This product includes software developed at' # NOTICE
+  end
 end
--- a/features/fixtures/pom.xml
+++ b/features/fixtures/pom.xml
@@ -15,5 +15,10 @@
       <version>4.13.1</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <version>3.12.0</version>
+    </dependency>
   </dependencies>
 </project>
--- a/lib/license_finder/package_managers/maven.rb
+++ b/lib/license_finder/package_managers/maven.rb
@@ -13,22 +13,20 @@ module LicenseFinder
     end
 
     def current_packages
+      # Generate a file "target/generated-resources/licenses.xml" that contains a list of
+      # dependencies including their groupId, artifactId, version and license (name, file, url).
+      # The license file downloaded this way, however, is a generic one without author information.
+      # This file also does not contain further information about the package like its name,
+      # description or website URL.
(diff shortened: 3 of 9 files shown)


xtreme-shane-lattanzio:
Hey @rhuitl! This looks great! I don't know maven that well but it looks like you kept the old way intact while adding the jar stuff.

As for adding a good test for this, I think you want to look at the testing_dsl file which creates tests projects to be tested on. You will need to create a project with a jar file in it and then run license finder on that new test project. You then add a case to the maven_spec.rb under features->package_managers. Does this make sense?

xtreme-shane-lattanzio:
Hey @rhuitl ! Is there any plan to continue with this? If we got the test in we will be ready to merge this. Let me know! Thanks!!

rhuitl:
Hi @xtreme-shane-lattanzio I added the test!

xtreme-shane-lattanzio:
Hey @rhuitl ! It looks like after merging this in there is a nil error on line 27 of the maven_dependency_finder for some cases:

```
LicenseFinder stderr:
/usr/share/rvm/gems/ruby-3.1.1/gems/license_finder-7.1.0/lib/license_finder/package_utils/maven_dependency_finder.rb:27:in `add_info_from_m2': undefined method `tr' for nil:NilClass (NoMethodError)

        .join(dep['groupId'].tr('.', '/'))
```

It looks like there is not a guarantee to have a group ID in the dep. Do you think there should be a check or something here?

rhuitl:
Hi @xtreme-shane-lattanzio do you have a package name for me so I can investigate? We could skip packages like this but it would be better to be able to locate the package archive.

xtreme-shane-lattanzio:
Hey @rhuitl It looks like this is happening whenever running on a package that has no dependencies. The package I was testing with is private but basically if you run `./mvnw org.codehaus.mojo:license-maven-plugin:download-licenses` to get a file that looks like

```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<licenseSummary>
  <dependencies/>
</licenseSummary>
```

Then you will have empty dependencies but then this

`Pathname.glob(@project_path.join('**', 'target', 'generated-resources', 'licenses.xml')).map(&:read).flat_map { |xml| XmlSimple.xml_in(xml, options)['dependencies'] }.each { |dep| print dep }`

command will still run leading it to look at values in the xml that do not exist. Note in this case, dep would come out as `{}`

rhuitl:
Does https://github.com/pivotal/LicenseFinder/pull/961 fix it for you, too?

Issue #927, opened 2022-07-15
Fail to detect "BSD 3-Clause"


taisei-watanabe:
I found that LicenseFinder fails to detect `"BSD 3-Clause"` as New BSD license (`BSD-3-Clause`). Adding `'BSD 3-Clause'` to `other_names` in [definition.rb](https://github.com/pivotal/LicenseFinder/blob/3c50484a55219543cd2947d28f685b70ab58f47f/lib/license_finder/license/definitions.rb#L258-L279) will solve the issue.

Note that `BSD-2-Clause` and `BSD-4-Clause` have `'BSD 2-Clause'` and `BSD 4-Clause` in `other_names`, respectively.

https://github.com/pivotal/LicenseFinder/blob/3c50484a55219543cd2947d28f685b70ab58f47f/lib/license_finder/license/definitions.rb#L258-L279

PR #928, opened 2022-07-20
Fail to detect "BSD 3-Clause"

taisei-watanabe:
## Issue

#927 

## What I did

- Add `'BSD 3-Clause'` in `other_names` of NewBSD license to solve a mis-recognition.


xtreme-shane-lattanzio:
@taisei-watanabe Good catch! Thanks for adding this in!

PR #936, opened 2022-08-25
[FIXED] - Yarn2 output parsing

xtreme-shane-lattanzio:


Kurt-von-Laven:
I was able to reproduce the issue this pull request addresses. If it's helpful to have a public Yarn v3 PnP repository to test this on, you are welcome to test against [ScribeMD/docker-cache](https://github.com/ScribeMD/docker-cache).

xtreme-shane-lattanzio:
Hey @Kurt-von-Laven ! Thanks for the sample repo. I tried to run both the master branch and this branch on the repo. Master returns 
```
LicenseFinder::NPM: is active
LicenseFinder::Yarn: is active

No dependencies recognized!
```

However, this branch returns
```
LicenseFinder::NPM: is active
LicenseFinder::Yarn: is active

@actions/cache, 3.0.0, MIT
@yarnpkg/sdks, 2.6.2, "Simplified BSD"
docker-cache, ., unknown
jest-junit, 14.0.0, "Apache 2.0"
```

Do you know what the expected output should be? I did notice one problem where it wasn't actually looping properly but I have now fixed this so the output for this repo is now

```
LicenseFinder::NPM: is active
LicenseFinder::Yarn: is active

@actions/cache, 3.0.0, MIT
@actions/core, 1.9.1, MIT
@fast-check/jest, 1.1.0, MIT
@jest/globals, 28.1.2, MIT
@jest/types, 28.1.1, MIT
@tsconfig/node16-strictest-esm, 1.0.3, MIT
@types/eslint, 8.4.5, MIT
@types/jest, 28.1.4, MIT
@types/node, 18.0.1, MIT
@types/prettier, 2.6.3, MIT
@vercel/ncc, 0.34.0, MIT
@yarnpkg/sdks, 2.6.2, "Simplified BSD"
docker-cache, ., unknown
eslint, 8.19.0, MIT
fast-check, 3.1.4, MIT
jest, 28.1.2, MIT
jest-junit, 14.0.0, "Apache 2.0"
prettier, 2.7.1, MIT
ts-jest, 28.0.5, MIT
ts-node, 10.8.2, MIT
typescript, 3, "Apache 2.0"
```

Please let me know if this output is more of what is expected!

Kurt-von-Laven:
I haven't hand-checked it at this stage, but that intuitively seems like a great start. The list seems to exclude transitive dependencies. Is that intentional? I would pass `--recursive` to yarn-plugin-licenses since my understanding is that companies are generally equally responsible for complying with the licenses of transitive dependencies. It's a little odd to me that docker-cache is considered a dependency of itself, but that seems like an issue with the Yarn plugin or possibly Yarn itself. In any case, LicenseFinder already offers excellent support for exceptions, so I personally wouldn't block on something minor like that; maybe there is some compelling rationale for that design I am unaware of.

xtreme-shane-lattanzio:
@Kurt-von-Laven I don't know if it is right or not to add the recursive flag so I will leave that to others. 

It sounds like this PR is good to go but I want to make sure it also solves the issue here: https://groups.google.com/g/license-finder/c/yaJeO1okF4c

Thanks again for your feedback!

Kurt-von-Laven:
Thanks for the update! I have to say I was quite excited when I found this pull request because I have been looking for a license compliance solution that supports Yarn PnP for over a year.

Regarding passing `--recursive`, I would argue that a license checker that only checks top-level dependencies is actually more dangerous than not using a license checker at all, because it checks only the tiny fraction of the dependency tree that would be most likely to be inspected manually while providing false reassurances of legal compliance. We work extensively with a wide variety of open-source projects, and almost none of them have a license compliance solution. We already found one case of an MIT-licensed project depending on a GPL'd project, and I suspect this is quite common since open-source projects typically lack for legal counseling and have little fear of being sued for a license violation as they rarely have any money. As [the LicenseFinder README mentions](https://github.com/pivotal/LicenseFinder/blob/d897f37d7e236682cf65dada8b516339df97ae9d/README.md#a-plea-to-package-authors-and-maintainers), some open-source projects lack license information altogether.

xtreme-shane-lattanzio:
@Kurt-von-Laven Just to try it out, I ran `yarn licenses list --json -R` on the repo you shared but it fails out with 
`Internal Error: ENOENT: no such file or directory, open '/workspace/docker-cache/.yarn/unplugged/fsevents-patch-3340e2eb10/node_modules/fsevents/package.json'`

I guess in order for this to work we would need to recursively run prepare commands on the project and then conditionally pass in the recursive flag if the license finder recursive option is used. Im gonna merge this as is and we can iterate on it if we want to!

Kurt-von-Laven:
Thanks for giving it a shot (and I'm pretty much always a fan of an iterative approach to open-source). What prepare commands do you have in mind here? The one I'm thinking of is `yarn install`, which always runs recursively.

Kurt-von-Laven:
Linking this to #936.

xtreme-shane-lattanzio:
@Kurt-von-Laven Well right now it runs `yarn install && yarn plugin import https://raw.githubusercontent.com/mhassan1/yarn-plugin-licenses/#{yarn_licenses_plugin_version}/bundles/@yarnpkg/plugin-licenses.js` if you pass `-p` into license finder. 

Then it runs `yarn licenses list --json` if you do not pass in the -p to license finder, it will just do the license command.

I tried to do the prepare command and then run `yarn licenses list --json -R` and I still got that error

Kurt-von-Laven:
Sounds like we are on the same page, but this means there's no prepare command that needs to be run recursively. Maybe the error is related to tophat/yarn-plugin-licenses#21 though? Note that that repository is a different Yarn license plugin than the one being used here. It performs the validation itself but only supports Yarn 3.0.0.

xtreme-shane-lattanzio:
What I meant was if the license finder recursive flag is used, it will actually go through each subdependency and run the prepare command in it. Then the license command should work with the recursive flag. It is possible that a different plugin is better. There is logic to use different versions for 2 and 3 while 1 doesn't even use a plugin

Kurt-von-Laven:
No, there's no reason to run `yarn install && yarn plugin import https://raw.githubusercontent.com/mhassan1/yarn-plugin-licenses/#{yarn_licenses_plugin_version}/bundles/@yarnpkg/plugin-licenses.js` on a per-dependency basis. `yarn install` is inherently recursive, and there's no benefit to importing a separate copy of the same plugin for each dependency. I believe the issue you encountered has nothing to do with transitive dependencies beyond the coincidence that fsevents happens to be a transitive dependency of docker-cache. fsevents is notable, because it's [MacOS-only](https://github.com/fsevents/fsevents/blob/66be5194c536a4a394a963ab5034696f9059fbab/package.json#L7-L9). Assuming you were on Linux or Windows, `yarn install` wouldn't have installed it, leading to the error you saw. If there is a better plugin out there to use, I would be delighted, but I haven't found it in a year of searching. Yarn v1 has native support for listing licenses, which was removed in Yarn v2. We may need support from Yarn to install all dependencies regardless of OS (e.g., `yarn install --all`), but I don't see it as a reason to refrain from checking transitive dependencies in LicenseFinder in the meantime for the benefit of projects that don't depend on OS-specific packages.

Kurt-von-Laven:
@xtreme-shane-lattanzio, would you be able to give your local change with `--recursive` a second try (and/or open a pull request containing it) now that ScribeMD/docker-cache#156 is merged?

Kurt-von-Laven:
I've opened #948 for this; please let me know if I've done anything obviously wrong as I am unable to test my change locally.

PR #950, opened 2022-10-18
PyPI: Raise on error 

dmcauslan:
### Changed
 - When `--prepare_no_fail` isn't set, raise errors from PyPi rather than suppressing them.


xtreme-shane-lattanzio:
Hey @dmcauslan ! Thanks for the PR. Looks good!

PR #952, opened 2022-10-24
Fix a bug writing to the decisions file when there is nested inheritance

mtoneil:
## Fixed

* Don't write decisions from the Parent to the Child when there is nested inheritance (i.e. Child inherits from Parent which inherits from Grandparent)


xtreme-shane-lattanzio:
Hey @mtoneil ! Thanks for fixing this!

PR #953, opened 2022-10-24
Add ability to specify package version when setting a manual license

mtoneil:
## Changed

* Allow `--version` option when setting a manual license, and respect the package version when deciding on a license

===========================

Currently, adding a license for a package applies it to all versions of that package. However, packages sometimes change their license between versions.

This change allows the `--version` option to be specified when invoking `license_finder licenses [add | remove]`. Similar to how approvals work, once licenses for specific versions of a given dependency are added, any licenses added for that dependency without the --version option are ignored.

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,12 +345,40 @@ you should manually research what the actual license is.  When you
 have established the real license, you can record it with:
 
 ```sh
-$ license_finder licenses add my_unknown_dependency MIT --homepage="www.unknown-code.org"
+$ license_finder licenses add my_unknown_dependency MIT
 ```
 
-This command would assign the MIT license to the dependency
-`my_unknown_dependency`. It will also set its homepage to `www.unknown-code.org`.
+This command would assign the MIT license to all versions of the dependency
+`my_unknown_dependency`. If you prefer, you could instead assign the license
+to only a specific version of the dependency:
 
+```sh
+$ license_finder licenses add my_unknown_dependency MIT --version=1.0.0
+```
+
+Please note that adding a license to a specific version of a dependency will 
+cause any licenses previously added to all versions of that dependency to be 
+forgotten. Similarly, adding a license to all versions of a dependency will 
+override any licenses previously added to specific versions of that dependency.
+
+There are several ways in which you can remove licenses that were previously
+added through the `licenses add` command:
+
+```sh
+# Removes all licenses from any version of the dependency
+$ license_finder licenses remove my_unknown_dependency
+
+# Removes just the MIT license from any version of the dependency
+$ license_finder licenses remove my_unknown_dependency MIT
+
+# Removes all licenses from only version 1.0.0 of the dependency
+# This has no effect if you had last added a license to all versions of the dependency
+$ license_finder licenses remove my_unknown_dependency --version=1.0.0
+
+# Removes just the MIT license from only version 1.0.0 of the dependency
+# This has no effect if you had last added a license to all versions of the dependency
+$ license_finder licenses remove my_unknown_dependency MIT --version=1.0.0
+```
 
 ### Adding Hidden Dependencies
 
--- a/features/features/configure/assign_licenses_spec.rb
+++ b/features/features/configure/assign_licenses_spec.rb
@@ -14,6 +14,7 @@ describe 'Manually Assigned Licenses' do
     gem = developer.create_gem 'mislicensed_dep', license: 'Unknown'
     project.depend_on gem
     developer.execute_command 'license_finder licenses add mislicensed_dep Known'
+    expect(developer).to be_seeing('The mislicensed_dep dependency has been marked as using Known license!')
 
     developer.run_license_finder
     expect(developer).not_to be_seeing_something_like /mislicensed_dep.*Unknown/
@@ -22,17 +23,33 @@ describe 'Manually Assigned Licenses' do
 
   specify 'can be removed, revealing the default license for a dependency' do
     project = developer.create_ruby_app
-    gem = developer.create_gem 'mislicensed_dep', license: 'Default'
-    project.depend_on gem
-    developer.execute_command 'license_finder licenses add mislicensed_dep Manual_license'
+    gem_a = developer.create_gem 'mislicensed_dep_a', license: 'Default'
+    gem_b = developer.create_gem 'mislicensed_dep_b', license: 'Default'
+    project.depend_on gem_a
+    project.depend_on gem_b
+    developer.execute_command 'license_finder licenses add mislicensed_dep_a Manual_license_a'
+    developer.execute_command 'license_finder licenses add mislicensed_dep_b Manual_license_b'
 
     developer.run_license_finder
-    expect(developer).to be_seeing_something_like /mislicensed_dep.*Manual_license/
+    expect(developer).to be_seeing_something_like /mislicensed_dep_a.*Manual_license_a/
+    expect(developer).to be_seeing_something_like /mislicensed_dep_b.*Manual_license_b/
 
-    developer.execute_command 'license_finder licenses remove mislicensed_dep Manual_license'
+    developer.execute_command 'license_finder licenses remove mislicensed_dep_a Manual_license_a'
+    expect(developer).to be_seeing('The dependency mislicensed_dep_a no longer has a manual license of Manual_license_a')
+
+    developer.execute_command 'license_finder licenses remove mislicensed_dep_b Incorrect_license'
+    expect(developer).to be_seeing('The dependency mislicensed_dep_b no longer has a manual license of Incorrect_license')
 
     developer.run_license_finder
-    expect(developer).to be_seeing_something_like /mislicensed_dep.*Default/
+    expect(developer).to be_seeing_something_like /mislicensed_dep_a.*Default/
+    expect(developer).to be_seeing_something_like /mislicensed_dep_b.*Manual_license_b/
+
+    developer.execute_command 'license_finder licenses remove mislicensed_dep_b'
+    expect(developer).to be_seeing('The dependency mislicensed_dep_b no longer has a manual license')
+
+    developer.run_license_finder
+    expect(developer).to be_seeing_something_like /mislicensed_dep_a.*Default/
+    expect(developer).to be_seeing_something_like /mislicensed_dep_b.*Default/
   end
 
   specify 'can be assigned and removed by package version' do
@@ -45,6 +62,7 @@ describe 'Manually Assigned Licenses' do
     developer.execute_command 'license_finder licenses add dep_a GPL --version=1.0.0'
     developer.execute_command 'license_finder licenses add dep_a MIT --version=2.0.0'
     developer.execute_command 'license_finder licenses add dep_a GPL --version=3.0.0'
+    expect(developer).to be_seeing('The dep_a dependency with version 3.0.0 has been marked as using GPL license!')
 
     developer.execute_command 'license_finder licenses add dep_b GPL --version=1.0.0'
     developer.execute_command 'license_finder licenses add dep_b GPL --version=2.0.0'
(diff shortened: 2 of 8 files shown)


xtreme-shane-lattanzio:
Hey @mtoneil ! This is good and something that has come up a few times in the past. 

I just want to make sure that existing behaviour wont break for people. I think I see all of the test cases but to be sure, the intent here is if in the past someone added a license without this new cli option, then someone added that same license with a version, that new version entry would take priority?

And if someone added one with the version today and then alater dded without any version, the single version is what would be shown?

I am curious about whether if someone adds an "all" case if that should take priority over the specific versions which I think is the opposite of what you put here. Or maybe someone can actually provide `--version all` on the cli and it would override?

What do you think? Im open to either but just want to cover our options. Thanks again for the PR!

mtoneil:
Hey @xtreme-shane-lattanzio! Thanks for taking a look at the PR.

> I just want to make sure that existing behaviour wont break for people ... if in the past someone added a license without this new cli option, then someone added that same license with a version, that new version entry would take priority?

That's correct. So I think from this point of view all existing users will have the same experience until they choose to add a license with the new `--version` option.

> And if someone added one with the version today and then alater dded without any version, the single version is what would be shown?

Also correct. Or rather, it'll only "know" about the license for that specific version of the dependency and would report unknown for any other version.

> I am curious about whether if someone adds an "all" case if that should take priority over the specific versions which I think is the opposite of what you put here. Or maybe someone can actually provide --version all on the cli and it would override?

I wondered about the best approach here, as well. I was primarily driven by trying to match the behavior of the `approvals` commands, although you can decide if that is a worthy endeavor. From looking at the code and manual testing, I concluded that the "all versions" approval (i.e. any approval given to a dependency without the `--version` option) would be ignored as soon as you add an approval for a specific version. This part makes sense to me. But I also found that it would ignore the "all versions" approval even if that took place _after_ the approval for a specific version. That does seem a bit unnatural as it feels unexpected for a new decision to be accepted but have no practical effect. I figure this was not intentional, but also not noticed because once you start approving specific versions then you'd be likely to keep doing that.

One small thing to note here, is due to the way I've implemented tracking the license decisions internally, if you have an "all versions" and version-specific licenses, and you then `remove` all the version-specific licenses, the "all versions" license will kick back in automatically. This is different from how removing approvals works, because `unapprove`'ing a dependency does not take a version and wipes out all approvals for that dependency (both "all approvals" and version-specific). So I'm also curious how you think the `remove` behavior for licenses should work (e.g. should it not accept `--version` and remove all licenses, or work the way I've implemented).

So I'm happy to implement in whatever way makes the most sense to you. I think if we say that later decisions always take precedence over earlier decisions (which is a sensible decision) then I do wonder if the `remove` should not take a version and just wipe all the licenses similar to approvals, to keep the implementation simple and not having to remember earlier overridden decisions that are now getting undone by a `remove`.

mtoneil:
@xtreme-shane-lattanzio  I was thinking about this a little more, and I think am landing on the following behavior, which is more or less what I was describing at the end of my last comment:

* When adding a license, more recent decisions always take precedence over earlier decisions. Example:
  1. User adds "all versions" license -> it applies to all versions
  2. User then adds license for version 1 -> now we only know about the license for version 1 and no other versions
  3. User then adds "all versions" license again -> it applies to all versions, overriding whatever was set for version 1
* When removing a license, it does not accept the version option and will remove all licenses set on that dependency, giving a clean slate to start over again if necessary

It's subtly different than how approvals works on the adding side, but makes a bit more sense to me and is easy to implement and document/explain. Thoughts?

xtreme-shane-lattanzio:
@mtoneil I think that is a good way to do it. It may be different than approvals but you are actually improving it. I do think that if you add an all case, it should take priority over previous entries. 

I also think removals bing a blanket removal is also good.

One last idea, I wonder if adding all will actually remove individual versions from the decisions file or if it will just append to the end. Similarly, if you add a specific version, will it remove any `all` entries or just append at the end and do logic. Maybe since removals will remove it all anyway, instead of appending, replacement makes more sense. I don't have a strong opinion on this but I was just thinking replacement can be cleaner since someone can manually edit the decision file incorrectly and change the order of things which would cause the output to change.

mtoneil:
@xtreme-shane-lattanzio Cool, I can whip something up for this soon.

> One last idea, I wonder if adding all will actually remove individual versions from the decisions file or if it will just append to the end. Similarly, if you add a specific version, will it remove any `all` entries or just append at the end and do logic. Maybe since removals will remove it all anyway, instead of appending, replacement makes more sense. I don't have a strong opinion on this but I was just thinking replacement can be cleaner since someone can manually edit the decision file incorrectly and change the order of things which would cause the output to change.

That's a good philosophical question about how the library should behave. Looking into the library a bit more, it seems there's a lot of pairs of "opposite" commands to do and undo something:
* approve / unapprove
* permit / unpermit
* restrict / unrestrict
* ignore / heed
* ignore_group / heed_group
* name_project / unname_project
* inherit_from / remove_inheritance

Only `remove_inheritance` removes existing decisions from the YML file, whereas all the rest append a new decision. (It definitely makes a lot of sense for inheritance to work this way, since the remote file may not exist after you remove the inheritance, plus the logic in the code would be a lot more complicated to remove the inherited rules once an uninherit decision is encountered.)

So it seems that the library largely works in append-only mode today. This does simplify the implementation a bit, and eliminates a class of nasty bugs where we accidentally mess up someone's YML file when undoing something (whereas if the bug is just in the in-memory logic, that can be patched in a subsequent release and made to be like it never happened at all). Also, it's officially documented in the README:

> ...keep in mind that each decision has an associated timestamp, and the decisions are processed top-to-bottom, with later decisions overwriting or appending to earlier decisions.

(perhaps a little misleading as I don't believe the timestamps matter, only the order in the file, but the spirit here is right)

So my personal preference would be to keep this append-only to maintain consistency with other similar commands, and to simplify the implementation. But as an official maintainer I would certainly defer to you!

mtoneil:
@xtreme-shane-lattanzio I'm remembering now why I kept `--version` in the `unlicense` command. `unlicense` accepts the license to remove, not just the package name. So e.g. you can do this today:

* `license foo_gem MIT` => knows about MIT
* `license foo_gem BSD` => knows about both MIT and BSD (I think implicitly it's OR'ing them)
* `unlicense foo_gem MIT` => knows only about BSD now

Frankly I'd really prefer if `unlicense` only accepted the package name and wiped everything. But that'd break compatibility. So I think in a world where `license` accepts `--version`, even if `unlicense` doesn't accept `--version` I'm not quite sure how I'd handle it if it also requires passing the specific license to remove. Unless it'd mean "remove this license from this dependency for any version, whether it be 'all versions' or specific versions".

Any thoughts on this? Feels a bit tricky.

xtreme-shane-lattanzio:
@mtoneil  Keeping append only is fine with me especially if the underlying library has that intent. We dont want to hack what it is trying to do. 

Okay so in general my main concern is backwards compatibility so whatever we can do to maintain that is my vote.

Assuming the library knows how to properly handle "all" being added after a version or vice versa, for new additions I think I would expect:
- license foo_gem MIT => Adds MIT at "all" versions for this gem
- license foo_gem MIT 1.0 => Adds MIT at only 1.0 for this gem
- license foo_gem 1.0 => Adds 1.0 and "all" licenses for the gem
- license foo_gem => Adds "all" versions and licenses for the gem

For removals, if you run unlicense, I think the cases would be:

- unlicense foo_gem MIT => Removes MIT at all versions for this gem
- unlicense foo_gem MIT 1.0 => Removes MIT at only 1.0 for this gem
- unlicense foo_gem 1.0 => Removes 1.0 and all licenses for the gem
- unlicense foo_gem => Removes all versions and licenses for the gem

It also sounds we might want to explicitly add "all" commands of some kind to use the `remove_inheritance` instead of other commands in case a user will need quick cleanup because with these version additions, I do worry about the file becoming large for no reason with duplicate entries since it only takes the latest.

Again at the end of the day, as long as we provide new value here while maintaining existing behaviour, Im happy. I appreciate you digging into the library to get the intent of it as well.

mtoneil:
Hey @xtreme-shane-lattanzio.

I think I like where you're going with these commands. However regarding these 2 "add" commands:
>* license foo_gem 1.0 => Adds 1.0 and "all" licenses for the gem
>* license foo_gem => Adds "all" versions and licenses for the gem

Do those make sense in the "add" context? I think for the "remove" (unlicense) context I can see how the license itself becomes optional (e.g. you're just saying "remove all licenses for this dependency, optionally for a specific version"). But when assigning the license, the license itself must be required, right?

> It also sounds we might want to explicitly add "all" commands of some kind to use the remove_inheritance instead of other commands in case a user will need quick cleanup because with these version additions, I do worry about the file becoming large for no reason with duplicate entries since it only takes the latest.

Do you have an example in mind of what that might look like? My gut is that since today the library doesn't make any effort to remove orphaned data, this is potentially a broader problem, and I'm not sure this change would exacerbate it in any meaningful way. For example you can already `approve` dependencies at specific versions, but that doesn't remove any earlier approvals from the file even if they no longer would have an effect. I almost wonder if it'd be better to have a generic command which produces a new file which is equivalent as far as outcomes but removes all orphaned decisions. But I also think having a "large" file doesn't seem like an issue in practice, unless license_finder actually becomes slow due to it. But for that to happen I'd think it'd have to be a quite large file. In practice my team is using lots of `permitted_licenses` and only needs to `approve` or `license` things here and there, so even at the version level I don't see this file getting meaningfully large. Users can always do some manual cleanups to remove orphaned data if they really want to, but I suspect that's not worth the effort and time for very little payoff. Anyway, let me know if I've misinterpreted your suggestion or if you have a different perspective.

xtreme-shane-lattanzio:
Hey @mtoneil ! Your point about adding requiring a license makes sense so lets just leave those cases alone. I was more thinking of streamlining things but I guess it does make sense to be more explicit in this case.

As for the remove all concept, similar to your suggestion of creating a new file I was meaning that there is a command to wipe out the existing entries for a given dependency in case a user wants to start over. For example picture a case like:

- someone adds foo_gem at version 1.0, 2.0, 3.0 and each with MIT
- foo_gem license gets updated to GPL
- user wants to remove all instances of foo_gem and add just foo_gem 3.0 with GPL

The way they would do this is with a single remove all command for the gem and then add as normal. I know if you added right now, it would take priority but I was thinking a cleanup is a good idea.

I guess I am going out of scope for this PR so Ill leave this alone as well and perhaps we will have a future enhancement. Either way I appreciate talking through this with you!

mtoneil:
Hey @xtreme-shane-lattanzio. I hear what you're saying, and agree it is best left to a possible future enhancement.

I will try to make some updates to this PR this week to sync up with what we've agreed upon. Thanks for the good discussion!

mtoneil:
@xtreme-shane-lattanzio I've pushed up a new commit that tries to implement what we settled on:
https://github.com/pivotal/LicenseFinder/pull/953/commits/5502b129553a5778347a83463708dd7546931626

I made a new class `ManualLicenses` because the logic was getting hairy and storing the licenses in 2 data structures was the easiest to implement and follow (1 hash for "all versions" licenses, and 1 hash for "specific versions" licenses). In that class I also added comments using the examples you provided to help indicate how it's intended to work and make clear the various use cases it supports. If you don't think the comments are appropriate to be there, let me know and I can remove them.

To summarize, the logic now works like this:
* When adding licenses, you can add to all versions or specific versions, but whichever you did most recently is what takes precedence and that's the "mode" you have to operate in until you approve the other way. For example, you say foo_gem is MIT for all versions, then later say foo_gem is GPL for version 1.0. Now it'll only know about version 1. And vice versa, if you started off saying foo_gem is GPL for v1, then you later say foo_gem is MIT for all versions, now it only knows about MIT for all versions. This isn't changed from my earlier commits, but wanted to be clear on how it works.
* When removing licenses, the license is now optional and the version is also optional. We'll remove the requested license (or all licenses, depending on whether the license was specified) from both "all versions" and "specific versions" when the version option is omitted, or only from "specific versions" when the version option is provided. This should line up with the bulleted examples you gave in your earlier comment, and I reflected them in the code as well.

I would like to do some more manual testing with the built gem locally before this gets merged, but wanted to push the changes up to get your feedback early.

xtreme-shane-lattanzio:
Hey @mtoneil ! This looks good to me. I dont mind comments either. It can be helpful to others in the future as well

The removal of all licenses is new now right? Since having a license was mandatory before? This is fine since it should not break backwards compatibility but we just need to be sure we document it.

mtoneil:
@xtreme-shane-lattanzio That's correct, the license being optional for `licenses remove` is new. And agreed that it does not break backwards compatibility.

As for documentation, I'm going to take a stab at adding some more detail into the Manual Intervention > Setting Licenses section of the README.

mtoneil:
@xtreme-shane-lattanzio Please have a look at these changes to the README, and let me know if anything is unclear or if I missed something: https://github.com/pivotal/LicenseFinder/pull/953/commits/3561c8436199bd1b2e97d13ad132fa93a51957eb

Also, I removed `--homepage="www.unknown-code.org"` from the example command because that option is in fact not supported by the `licenses` commands.

mtoneil:
@xtreme-shane-lattanzio I did a bunch of manual testing on the updated gem locally on of my own repos and satisfied it's working as desired per the new README description

xtreme-shane-lattanzio:
Hey @mtoneil ! Thanks for finishing this up! Looks good! I was surprised to see the functionality to remove all versions and licenses of a dependency but I think we just had a misunderstanding. I meant to have a full remove command available but looking back I think it looked like I was only asking for an add all scenario.

Either. way this looks good and will merge when tests go through. Thanks again!!

mtoneil:
Thanks @xtreme-shane-lattanzio! Any plans to cut a new release soon?

xtreme-shane-lattanzio:
@mtoneil Yep! Just waiting for the pnpm MR to finish up and then will cut one :)

PR #959, opened 2022-12-15
npm packages: do not crash if an array is given for author

tomcornelis:
I encountered a crash 
```
lib/license_finder/packages/npm_package.rb:92:in `author_name': no implicit conversion of String into Integer (TypeError)
```
which was caused by author being an array instead of string or dictionary in the following package.json:
https://github.com/romulomachado/ember-cli-string-helpers/blob/master/package.json#L55-L59

This PR makes sure we do not crash on such cases.

### DIFF
--- a/lib/license_finder/packages/npm_package.rb
+++ b/lib/license_finder/packages/npm_package.rb
@@ -80,16 +80,22 @@ module LicenseFinder
 
     def author_names
       names = []
-      names.push(author_name(@json['author'])) unless @json['author'].nil?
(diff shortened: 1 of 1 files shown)


xtreme-shane-lattanzio:
Hey @tomcornelis ! Thanks for the PR. It looks like someone else made an issue and gave an example where the author object type is checked in the calling method and then parsed out before grabbing the name itself. I think it is a cleaner approach. Could you update it to that?

PR #962, opened 2023-01-12
Maven: add test to verify that an empty dependencies section is handled gracefully

rhuitl:
I had planned to include a test in #961 but wasn't able to finish it yesterday, so here it is :-)


xtreme-shane-lattanzio:
@rhuitl Thanks for that! Sorry I was hasty merging the other one in. Since it was simple, I just wanted to merge it and have our pipelines handle it. They are still going but if we are all good, Ill merge this one in as well :)

PR #967, opened 2023-02-08
Clean up the Dockerfile a bit.

rhuitl:
Here are some smaller cleanups to the Dockerfile. I've been cleaning up a few other Dockerfiles in the past and I saw this one could need some love, as well :-)

* Remove APT files after each RUN (see also https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get)
* Use the more modern `apt` command instead of `apt-get` for brevity
* Remove unneeded `sudo` calls

The first point might lead to a tiny reduction in the image size because we're clearing the APT files after each run. On the other hand, the build time might go up a bit because of all the repeated `apt update` calls. If you want, I can also try to combine all installs into a single RUN line. But maybe you have different ideas how to reduce the image size.

I'll post the size before and after once the CI/CD has built the image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,28 +14,28 @@ ENV MIX_VERSION 2.0
 ENV COMPOSER_ALLOW_SUPERUSER 1
 
 # programs needed for building
-RUN apt-get update && apt-get install -y \
+RUN apt -q update && apt install -y \
     build-essential \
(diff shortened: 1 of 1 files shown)


xtreme-shane-lattanzio:
Thanks @rhuitl !

PR #977, opened 2023-05-01
Use Ubuntu 22.04 (Jammy) for Docker image

rhuitl:
@xtreme-shane-lattanzio FYI, not sure if all the tests will pass.

Update:
* Python2 is no longer available. Do we need it? I changed the default in the code to use "3" and that seems to make the tests pass.
* The Go tool `dep` is deprecated since 2020 and not available on Jammy. See also https://github.com/golang/dep#dep. Do we need it?

Somewhat related:
In general, it might be worth considering not to build one gigantic Docker image that has all possible package managers and supporting tools and frameworks in it, but rather make LicenseFinder itself use smaller Docker images to perform the scans.
This way, it could use the most up-to-date, unmodified upstream images (e.g. https://hub.docker.com/_/maven, https://hub.docker.com/_/python, etc.) or even ancient ones with legacy software. Further, LicenseFinder users would only have to pull (or wait for LF to pull) the Docker images they actually need.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,14 @@
-FROM ubuntu:bionic
+FROM ubuntu:jammy
 
 WORKDIR /tmp
 
 # Versioning
-ENV PIP_INSTALL_VERSION 19.0.2
 ENV PIP3_INSTALL_VERSION 20.0.2
 ENV GO_LANG_VERSION 1.17.13
 ENV MAVEN_VERSION 3.6.0
 ENV SBT_VERSION 1.3.3
 ENV GRADLE_VERSION 5.6.4
 ENV RUBY_VERSION 3.2.2
-ENV MIX_VERSION 2.0
 ENV COMPOSER_ALLOW_SUPERUSER 1
 
 # programs needed for building
@@ -29,7 +27,7 @@ RUN add-apt-repository ppa:git-core/ppa && \
     apt -q update && apt install -y git && rm -rf /var/lib/apt/lists/*
 
 # install nodejs
-RUN curl -sL https://deb.nodesource.com/setup_17.x | bash - && \
+RUN curl -sL https://deb.nodesource.com/setup_18.x | bash - && \
     apt -q update && apt install -y nodejs && rm -rf /var/lib/apt/lists/*
 
 # install yarn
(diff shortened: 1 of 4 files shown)


xtreme-shane-lattanzio:
Hey @rhuitl ! I am looking into some things internally to ensure that we can drop support for dep and python2. We build other images off of the LF image so we will need to bump them as well and then change tests for those apps to also not use these old tools.  I think it should be okay but need to double check. This is a pretty big change but it does make sense to do and then release an 8.0.

In the mean time, there are some tests failing for the default python and pip versions. It is also not a big deal, but I'm not sure if there was a functional reason why you moved the flutter stuff?

As for breaking down the image, that would require a lot of changes for us so I would still want a single big one but I am not against adding more dockerfiles in addition to satisfy what you are talking about.

Thanks again for adding this!

xtreme-shane-lattanzio:
The tests that are failing are pretty trivial to fix. It is just because pip and python are using 3 instead of 2 so now the expect commands are saying for example:
``` 
expected: ("python /tmp/build/aa1bbb9d/LicenseFinder/bin/license_finder_pip.py some-file.txt")
got: ("python3 /tmp/build/aa1bbb9d/LicenseFinder/bin/license_finder_pip.py some-file.txt")
```

This is in a few places but you can just make the change and push and see the pipeline go green. The integration tests weren't run because the unit tests were red but we can tackle that next.

The only other failure is `it { expect(Scanner::PACKAGE_MANAGERS).to include described_class }` because you removed Dep from the list so that one should be an easy fix as well.

Again, thanks for your work on this!

rhuitl:
Ah thanks, and about the "Dep", where does it get the described_class from? Do I have to remove all files related to "Dep"?

xtreme-shane-lattanzio:
Unfortunately if I remember correctly, the described class in this case is anything that extends the PackageManager class. So you would either need to remove the extend, comment out the class, or remove it. That would definitely case some integration tests to fail which would be in the testing_dsl.rb. There are some Dep test projects in there. 

I forget if there is more than that that will cause trouble. I know there are dep fixtures but if you remove the link to package managers, I imagine they will never be looked at.

xtreme-shane-lattanzio:
Hey @rhuitl ! The bionic deprecation seems to be causing new issues. I am wondering if you are able to do the dep removal? Let me know! I am working internally here to see if anything else needs to be done for our tools before we merge and release this. Thanks!!

xtreme-shane-lattanzio:
@rhuitl FYI I opened up https://github.com/pivotal/LicenseFinder/pull/992 so that I can tinker around with this as well and hopefully get the upgrade through!

rhuitl:
> @rhuitl FYI I opened up #992 so that I can tinker around with this as well and hopefully get the upgrade through!

Hi Shane, that's great! I hope it'll be easy, but as you said, it's probably mostly about deciding whether we can remove some older package managers that are no longer available and then remove all related code to make the tests happy.

In the past weeks, I've been busy with non-tech stuff and didn't have time to look into this further.

PR #986, opened 2023-06-21
Support of Conan v1.6 and v2

melter2m:
## Motivation
Conan 1.6 extracts licenses to different path than expected which causes License Finder to crash.

### Reason
File mask that is used to search licenses looks as follows:

`#{project_path}/licenses/#{name}/**/LICENSE*`

Conan 1.6 creates license files under `#{project_path}/licenses/#{name}/licenses/LICENSE*` and first file that suites is directory

### Solution

Iterate over the files that suite mentioned above mask until we find not directory

### DIFF
--- a/lib/license_finder/package_managers/conan.rb
+++ b/lib/license_finder/package_managers/conan.rb
@@ -10,15 +10,13 @@ module LicenseFinder
     end
 
     def license_file_is_good?(license_file_path)
-      license_file_path != nil && File.file?(license_file_path)
+      !license_file_path.nil? && File.file?(license_file_path)
     end
 
     def license_file(project_path, name)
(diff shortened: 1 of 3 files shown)


xtreme-shane-lattanzio:
Hey @melter2m ! Thanks for the PR! We are working on getting the jammy image upgrade since it is causing all kinds of issues right now. Once thats in, hopefully the pipelines on this PR pass and we will merge it in!!

xtreme-shane-lattanzio:
Hey @melter2m ! It looks like these changes do actually fail a test. Can you take a look and fix so we can have this merged in?

melter2m:
> 

Fixed

melter2m:
> Hey @melter2m ! It looks like these changes do actually fail a test. Can you take a look and fix so we can have this merged in?

fixed

Issue #990, opened 2023-07-08
Ubuntu PHP PPA has dropped support for Bionic


swarren12:
Currently, the License Finder Docker image is built off `ubuntu:bionic`.

As part of the construction of this image, the [Ubuntu PHP PPA](http://ppa.launchpad.net/ondrej/php/ubuntu) is added and, from this PPA, `php7.4` is installed.

As of July 2023, Bionic is officially [unsupported](https://wiki.ubuntu.com/Releases) and so the PPA has dropped support for it.

> Co-installable PHP versions: PHP 5.6, PHP 7.x, PHP 8.x and most requested extensions are included. Only Supported Versions of PHP (http://php.net/supported-versions.php) for Supported Ubuntu Releases (https://wiki.ubuntu.com/Releases) are provided. Don't ask for end-of-life PHP versions or Ubuntu release, they won't be provided.

As a result, currently the License Finder Docker image cannot be built, as the `php7.4` package is unresolvable.

Exact error when trying to build:

```
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package php7.4-cli
E: Couldn't find any package by glob 'php7.4-cli'
E: Couldn't find any package by regex 'php7.4-cli'
ERROR: Invalid installer checksum
[...]

PR #991, opened 2023-07-09
Ensure licenses are recognised by `spdx_id`

### LINKED ISSUE #984: [BUG]  'permitted_licenses add' not working properly for all licenses using spdx-id format.  LGPL and Python for sure.
In my workflow I have a step that contains a for loop that iterates over a list of licenses and adds them as permitted.
```
- name: 'Add Permitted Licenses'  #This is a list of Permitted Licenses, add comma separated licenses here to prevent them from 'failing' in the report
        run: |
            echo "Configuring Permitted Licenses"
            licenses="MIT,Apache-2.0,ISC,Ruby,BSD-2-Clause,BSD-3-Clause,BSD-4-Clause,WTFPL,0BSD,CC0-1.0,PSF-2.0,MPL-2.0,LGPL-2.0-only,LGPL-3.0-only,LGPL-2.1,LGPL,Python"
            for license in ${licenses//,/ }
            do
                license_finder permitted_licenses add "$license"
            done
        shell: bash
```

I can only confirm that for sure the PSF-2.0 and LGPL-3.0-only do not work properly.   These are the SPDX-id name format of the license names as found in the definitions file in this repo. LicenseFinder/lib/license_finder/license/definisions.rb

We kept getting packages 'failing' and reporting those licenses as the ones detected.  It was only after adding the 'short name' of these licenses at the end, 'LGPL' and 'Python',  did these packages detected with those licenses start passing.

I know this is not the intended behavior of the command because I use the spdx-id format of the Apache 2.0 license, Apache-2.0, in the same list and those packages pass, no problem.

For my purposes, the fix is to add the short names of these licenses as well, and it works, but wanted to point this out to you all and get it in the issues in case others are having trouble with this and come here for help.

swarren12:
The recognition of licenses named by their SPDX ID is currently inconsistent; for some licenses the ID is included as an alternative name, but for some it is not. This leads to inconsistent behaviour, as seen in issue #984.

Changing the matching logic to check the SPDX ID will make this more consistent and intuitive.

This fixes #984

### DIFF
--- a/lib/license_finder/license.rb
+++ b/lib/license_finder/license.rb
@@ -85,7 +85,7 @@ module LicenseFinder
     attr_reader :short_name, :pretty_name, :other_names, :spdx_id, :matcher
 
     def names
-      ([short_name, pretty_name] + other_names).uniq
+      ([short_name, pretty_name, spdx_id] + other_names).uniq
     end
   end
 
--- a/lib/license_finder/license/definitions.rb
+++ b/lib/license_finder/license/definitions.rb
@@ -46,12 +46,9 @@ module LicenseFinder
       def apache1_1
         License.new(
           short_name: 'Apache1_1',
-          pretty_name: 'Apache 1.1',
           spdx_id: 'Apache-1.1',
-          other_names: [
-            'Apache-1.1',
-            'The Apache Software License, Version 1.1'
-          ],
+          pretty_name: 'Apache 1.1',
+          other_names: ['Apache Software License, Version 1.1'],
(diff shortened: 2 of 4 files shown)


swarren12:
I assume that CI is failing due to https://github.com/pivotal/LicenseFinder/issues/990.

xtreme-shane-lattanzio:
Hey @swarren12 ! Thanks for the PR! Yes the pipelines are failing because of docker base images. We are working on merging the jammy upgrade and once thats done, we will see if the pipelines for this PR pass and will merge in!!

PR #1011, opened 2023-12-21
Speed up Cargo license finder

seanlinsley:
The current implementation uses a recursive `**` glob, which is very slow on our application. `time ls ~/.cargo/registry/src/**/ahash-0.8.6` for example takes 1.8 seconds for a single package.

Changing it to use `*` improves `bundle exec license_finder` runtime from several minutes (at least) to 28 seconds.

I don't think the `**` was needed: the path includes a folder for every Cargo registry, immediately followed by a folder for each crate.

### DIFF
--- a/lib/license_finder/package_managers/cargo.rb
+++ b/lib/license_finder/package_managers/cargo.rb
@@ -6,7 +6,7 @@ module LicenseFinder
   class Cargo < PackageManager
     def current_packages
       cargo_output.map do |package|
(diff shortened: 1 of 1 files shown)


seanlinsley:
I don't believe it's necessary to sign the CLA in this case because this is considered an [obvious fix](https://cla.pivotal.io/about#obvious-fixes). Let me know if it is needed in this case.

PR #1013, opened 2024-01-04
fix: For Ruby 3.3 compatibility, specify racc dependency in gemspec

faisal:
This PR adds racc to the gemspec, because [Ruby 3.3 changes](https://bugs.ruby-lang.org/issues/19702) mean that without it we get:

```
~/test (+main) $ cat Gemfile 
# frozen_string_literal: true

source "https://rubygems.org"
~/test (+main) $ license_finder
<internal:/opt/homebrew/Cellar/ruby/3.3.0/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:127:in `require': cannot load such file -- racc/parser.rb (LoadError)
	from <internal:/opt/homebrew/Cellar/ruby/3.3.0/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:127:in `require'
	from /opt/homebrew/lib/ruby/gems/3.3.0/gems/tomlrb-2.0.3/lib/tomlrb/generated_parser.rb:7:in `<top (required)>'
	from <internal:/opt/homebrew/Cellar/ruby/3.3.0/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:127:in `require'
	from <internal:/opt/homebrew/Cellar/ruby/3.3.0/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:127:in `require'
	from /opt/homebrew/lib/ruby/gems/3.3.0/gems/tomlrb-2.0.3/lib/tomlrb/parser.rb:1:in `<top (required)>'
	from <internal:/opt/homebrew/Cellar/ruby/3.3.0/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:127:in `require'
	from <internal:/opt/homebrew/Cellar/ruby/3.3.0/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:127:in `require'
	from /opt/homebrew/lib/ruby/gems/3.3.0/gems/tomlrb-2.0.3/lib/tomlrb.rb:9:in `<top (required)>'
	from <internal:/opt/homebrew/Cellar/ruby/3.3.0/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:127:in `require'
	from <internal:/opt/homebrew/Cellar/ruby/3.3.0/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:127:in `require'
[...]

PR #1014, opened 2024-01-11
Test on Ruby 3.3.0

benlangfeld:
In the hopes that it exposes https://github.com/pivotal/LicenseFinder/pull/1013


xtreme-shane-lattanzio:
@benlangfeld Ill add 3.3.x support to the pipeline. There are a few things missing from this so Ill make sure pipelines are added and they pass

PR #1016, opened 2024-01-14
[Fixed] Yarn 2 not working for recursive projects

### LINKED ISSUE #989: Yarn 2 does not work for recursive projects
It seems that Yarn (2) doesn't work with recursive scanning.

**Expected Behaviour**: Licenses for `yarn` dependencies are detected when running a command such as `license_finder report --recursive ....` with a project using `yarn` version `2.4.3`;
**Actual Behaviour**: Licenses for `yarn` dependencies are ignored;
**Steps to reproduce**: See https://github.com/swarren12/pivotal-license-finder-yarn2-bug

**Other information:**

It seems that the Docker image for 7.1.0 ships with `yarn` version `1.22.19`; however, using a `.yarnrc.yml` file this can be overridden. When using `yarn` version `2.4.3`, it seems Yarn licenses aren't detected.

From what I can see, this is because the `yarn` command is executed from the root directory (e.g. `/scan` in the Docker image). For `yarn` version 1, an extra argument is passed (`--cwd`) but this is omitted for version 2. In `yarn.rb`, there is at [least one occurrence of changing the working directory](https://github.com/pivotal/LicenseFinder/blob/master/lib/license_finder/package_managers/yarn.rb#L40C33-L40C78) (`Dir.chdir(project_path) { Cmd.run(prep_cmd) }`) but other invocations of `yarn` are missing this.

I believe the easiest fix would be to wrap all `Cmd.run()` calls in `yarn.rb` with a `Dir.chdir() {}` block, but I don't know if this would have a knock-on impact on other versions.

swarren12:
An attempt to fix #989 by making Yarn 2 run in the sub-project directory.


xtreme-shane-lattanzio:
LGTM!

PR #1037, opened 2024-04-13
Avoid using OpenStruct where Struct is enough

amatsuda:
On PR #1036, I saw a CI failure that is unrelated to its changeset.
https://norsk.cf-app.com/builds/446820782

The error `NameError: uninitialized constant LicenseFinder::OpenStruct` implies that we should explicitly bundle "ostruct" gem, but in this case I'd suggest to simply use Struct instead of OpenStruct.

### DIFF
--- a/spec/lib/license_finder/decisions_spec.rb
+++ b/spec/lib/license_finder/decisions_spec.rb
@@ -418,7 +418,7 @@ module LicenseFinder
       end
 
       it 'inheritates rules from gem decision file' do
-        gem_spec = OpenStruct.new(gem_dir: 'gem-name')
+        gem_spec = Struct.new(:gem_dir).new('gem-name')
         allow(Gem::Specification).to receive(:find_by_name).with('gem-name').and_return(gem_spec)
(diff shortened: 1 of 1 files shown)


amatsuda:
@pivotal-cla This is an Obvious Fix

amatsuda:
Ruby 2.6 CI fails https://norsk.cf-app.com/builds/446825268 due to ba4910e02d77ed7047a4d6dfaa8b690bfbc3846b.
Perhaps you'd better either revert this or drop Ruby < 2.7 support.

xtreme-shane-lattanzio:
@amatsuda Thanks for the PR. I saw this after but added the fix by explicitly requiring openstruct

amatsuda:
@xtreme-shane-lattanzio No problem, I'm totally fine with that. But maybe I should have wrote in the PR description that use of ostruct is now mildly warned due to performance reason https://github.com/ruby/ostruct/pull/56
and that's why I suggested to switch to Struct rather than simply adding a `require` statement.

Issue #1047, opened 2024-09-26
Add `-buildvcs=false` to Go scanning


emmahsax:
Would you be willing to add `-buildvcs=false` to the Go scanning command? I don't think we really need to get the VCS history as part of the license finder.

I am running this license finder in GitHub Actions as part of our CI/CD. But I keep getting this issue:
```
/usr/local/bundle/gems/license_finder-7.2.1/lib/license_finder/package_managers/go_modules.rb:63:in `block in packages_info': Command 'GO111MODULE=on go list -mod=readonly -deps -f '{{ if and (.DepOnly) (.Module) (not .Standard) }}{{ $mod := (or .Module.Replace .Module) }}{{ $mod.Path }},{{ $mod.Version }},{{ or $mod.Dir .Dir }}{{ end }}' ./...' failed to execute (RuntimeError)
	from /usr/local/bundle/gems/license_finder-7.2.1/lib/license_finder/package_managers/go_modules.rb:32:in `chdir'
	from /usr/local/bundle/gems/license_finder-7.2.1/lib/license_finder/package_managers/go_modules.rb:32:in `packages_info'
	from /usr/local/bundle/gems/license_finder-7.2.1/lib/license_finder/package_managers/go_modules.rb:20:in `current_packages'
	from /usr/local/bundle/gems/license_finder-7.2.1/lib/license_finder/package_manager.rb:105:in `current_packages_with_relations'
	from /usr/local/bundle/gems/license_finder-7.2.1/lib/license_finder/scanner.rb:42:in `each'
	from /usr/local/bundle/gems/license_finder-7.2.1/lib/license_finder/scanner.rb:42:in `flat_map'
	from /usr/local/bundle/gems/license_finder-7.2.1/lib/license_finder/scanner.rb:42:in `active_packages'
	from /usr/local/bundle/gems/license_finder-7.2.1/lib/license_finder/core.rb:84:in `current_packages'
	from /usr/local/bundle/gems/license_finder-7.2.1/lib/license_finder/core.rb:79:in `decision_applier'
	from /usr/local/lib/ruby/3.3.0/forwardable.rb:234:in `any_packages?'
	from /usr/local/bundle/gems/license_finder-7.2.1/lib/license_finder/license_aggregator.rb:17:in `block in any_packages?'
	from /usr/local/bundle/gems/license_finder-7.2.1/lib/license_finder/license_aggregator.rb:15:in `map'
[...]

PR #1048, opened 2024-09-27
Add `-buildvcs=false` to the `go list` command

emmahsax:
This will resolve the issues seen in https://github.com/pivotal/LicenseFinder/issues/1047, and has been tested with Go version 1.22.1.


emmahsax:
I'm not sure why the concourse-ci tests aren't running... it seems like maybe this repository has been disconnected from concourse-ci lately.

However, it also seems like maybe this repository is like half-archived and won't be updated any more at all, since nobody's really touched it since May 2024.

Dockerfile
```diff
@@ -4,7 +4,7 @@ WORKDIR /tmp
 
 # Versioning
 ENV PIP3_INSTALL_VERSION 20.0.2
-ENV GO_LANG_VERSION 1.17.13
+ENV GO_LANG_VERSION 1.18.10
```
emmahsax, line 7: This was my attempt to try to fix the concourse-ci tests, which may have been failing due to older versions of Go not supporting the `buildvcs` flag. Unfortunately, I'm not able to verify if this helps or not due to the concourse-ci tests no longer running on this PR.

PR #1058, opened 2025-05-12
lazy loading licensing

rurkss:
Optimize license_finder behavior for package by skipping unnecessary activations when license info is already resolved via decisions or spec

### Motivation
In Yarn v1, packages often include numerous transitive dependencies that get placed within the node_modules folder.  license_finder attempts a deep scan of the file system, looking for license-related files (e.g., LICENSE, COPYING, README) in these nested folders.

This scanning process is especially expensive for deeply nested packages and can take up to 10 seconds per package in the worst case. By skipping the activation step when license information has already been resolved via decisions or spec metadata, we significantly reduce unnecessary file system traversal and improve overall performance.

### DIFF
--- a/lib/license_finder/package.rb
+++ b/lib/license_finder/package.rb
@@ -18,7 +18,7 @@ module LicenseFinder
   #   the constructor options
   # - otherwise, override #licenses_from_spec or #license_files
   class Package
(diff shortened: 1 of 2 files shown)


wadewinningham:
Any update on this getting merged?

PR #1061, opened 2025-07-31
Add logger dependency to prepare for ruby 3.5.0

ylecuyer:
Add logger dependency to get ride of incomming deprecation warning of ruby 3.5.0

```
bundle/jruby/3.4.0/gems/license_finder-7.2.1/lib/license_finder/core.rb:5: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
```


gtupoto7-collab (approved):
Fg

